PR #26, opened 2020-06-01
base of bakery

guillaume-durville:
Hello Simon, 

Don't know how you want to do it, I was absolutely lazy, didn't rebase anything here and commented extensively in commit.

So what you have to do is place of course in your local folders : 
- OMD_ROOT/local/share/check_mk/agents/bakery/robotmk

The robotmk plugin MUST exist at the right place for the baking to work.

I have added logging to the bakery rule that logs to the usual place, here you can see the variables as things are baked.

I managed to bake an agent and get results, only formatting was bad inside because of pyyaml as far as I can see, but of course that can be correct.

Note that I import some so far not called base check_mk module in your original file to make the bakery rule work in the same file you did service discovery.

I let open the possibility of windows & also of a local configuration file/plugin as you can see with this fictitious robotmk.vbs !

I mostly looked at mk_oracle to guess how to build the GUI, and wish I could use AbsoluteDirname object from agent_bakery.p, but for that I would have needed another import and I was lazy. Please note that this seems interesting because it's used by the agent to learn about path in general. I don't know how usable that is, still.

So : 
- baking : check, 
- rule parameter : check, works with defaults
- creating local yaml file : uhhhhh, not yet

So a lot more to go for me, but I had a lot of fun so far :)

I don't know where to put the doc in particular, or how to discuss inside a wiki or anything. Let me know.

Of course, please tell me if there is anything I have not seen, used wrong, etc.

I feel like this is very little work compared to your effort so far, bravo again !


simonmeggle:
Hi Guillaume,

you are great, thanks so far for your work!
I will look into the code asap!
Kind regards, Simon


Guillaume Durville <notifications@github.com> schrieb am Mo., 1. Juni 2020,
17:57:

guillaume-durville:
Hey Simon,

Thanks, got a lot left to do but I felt happy understanding the pieces
today :)

I'll continue to improve what I already sent.

See you soon !

Guillaume DURVILLE


Le lun. 1 juin 2020 à 20:34, Simon Meggle <notifications@github.com> a
écrit :

simonmeggle:
Regarding robotmk.vbs: I think it won't be necessary to rewrite the whole
plugin in VBS. As we already have Python3 on the E2E-testing nodes
(beacuse of the Robot Framework installation), we have all we need.
All we have to do is to handle some OS peculiarities.

>

guillaume-durville:
Hey Simon,

For sure, it was just to showcase how they handle it in the bakery when
they need it.

A lot of stuff that I put in the pull request is what I would call
superfluous :
- the logging is not for prod
- the vbs par is just for show
- the variables in conf are just to demonstrate how they do it elsewhere
- the screenshot itself is just to have a vision of what kind of object
reaches the bakery

Honestly most of it is the product of me having to teach myself how the
bakery works.

We can remove & replace most of it.

Guillaume DURVILLE


Le jeu. 4 juin 2020 à 10:10, Simon Meggle <notifications@github.com> a
écrit :

PR #34, opened 2020-06-27
Added new options, safer yaml dump, stuck with tuple/unpacking logic …

guillaume-durville:
…& global data structure here, not sure. Regex check to be done in bakery file but implication is that baking can fail on it after user input (as we discussed). Don't want it to slow you down so I'm pushing what I have which is not much but I think I confused myself trying to integrate the new tuple/dict structure in there.

### DIFF
--- a/agents/bakery/robotmk
+++ b/agents/bakery/robotmk
@@ -4,117 +4,56 @@
 import cmk.utils.paths
 import logging
 import yaml
-import os
-
-#this is purely a debug feature
-logger = logging.getLogger('agent_bakery')
-logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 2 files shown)


simonmeggle:
Hi Guillaume,
thanks a lot for your contribution. I'll review this on Monday!
Regards, Simon

______________________________________________



*elabit* | *Simon Meggle*
IT-Monitoring und -Automatisierung
Gewerbering 3 | D-86504 Merching
simon.meggle@elabit.de | www.elabit.de


Am Sa., 27. Juni 2020 um 10:35 Uhr schrieb Guillaume Durville <
notifications@github.com>:

PR #237, opened 2023-07-04
Add --disallow-any-expr to mypy

jherbel:


SoloJacobs (commented):
Looks good!

SoloJacobs (dismissed):

SoloJacobs:
I looked a bit into disallow-any-expr, and it appears to me, that it will be quite a lot of hassle. We use it in agent_kube, but that project does very few dependencies. However, those tend to generate a lot of type errors, especially cmk/gui/. Maybe some other mypy flags are more useful to us.

PR #238, opened 2023-07-04
Avoid locking Pipfile when executing GitHub CI

jherbel:


SoloJacobs (dismissed):
This is an improvement, thus oti: Is the underlying problem that we have checkmk/Pipfile as a make dependency? The timestamp of this file does not reflect whether the ./pipenv lock should be triggered. With this change, will the ci trigger a ./pipenv lock if we update the checkmk submodule?

jherbel:
As discussed: I removed the Makefile to decouple locking and building the venv from running commands inside the venv.

SoloJacobs (approved):

PR #239, opened 2023-07-05
Add a check if all CI steps succeeded to GitHub actions CI

jherbel:
Apparently, the GitHub branch protection settings don't offer the option to enforce that all jobs have to succeed. Instead, each job has to be specifically configured. If a new job is added, it has to be added manually in the branch protection settings.

To avoid this manual maintenance, we use the solution offered by https://github.com/re-actors/alls-green.


simonmeggle (approved):
Looks good

PR #240, opened 2023-07-05
Improve caching in GitHub actions

jherbel:
Instead of caching the pip and pipenv caches, we cache the virtual environment.


SoloJacobs (approved):
LGTM

PR #245, opened 2023-07-07
Git hook for branch name validation

jherbel:


Makefile
```diff
@@ -0,0 +1,27 @@
+.PHONY: setup
+
+setup:
+	. ./python_version; \
```
SoloJacobs, line 4: I would have put this in a shell script, but oti.

.github/workflows/ci.yaml
```diff
@@ -34,26 +34,15 @@ jobs:
...
-        # librrd-dev: Needed for building rrdtool python bindings.
-        # libldap2-dev: Needed for building python-ldap.
-        # libsasl2-dev: Needed for building python-ldap.
-        # libkrb5-dev: Needed for building pykerberos.
-        sudo apt-get update
-        sudo apt-get install librrd-dev libldap2-dev libsasl2-dev libkrb5-dev
-        pip3 install pipenv
-    
-    - name: Cache virtual environment
-      id: cache-venv
```
SoloJacobs, line 48: I'm all for deleting caching steps, which are not needed. But I am bit a surprised it is done in this commit.

SoloJacobs (commented):
LGTM, but 2 comments

.githooks/pre-commit
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+VALID_BRANCH_REGEX="^(master$|dev\/.+)"
+
+if [[ ! $(git rev-parse --abbrev-ref HEAD) =~ $VALID_BRANCH_REGEX ]]
```
SoloJacobs, line 5: This will cause problems in detached head state :( Thus, it makes using
```
$ git commit --amend
```
during interactive rebase impossible.

SoloJacobs: We should probably use a pre-push hook instead.

SoloJacobs: This should do the trick
```
$ mv .githooks/pre-commit .githooks/pre-push
```
Works fine for me locally.

racicLuka (approved):
Looks good to me

PR #247, opened 2023-07-10
Cleanup pipenv and ci scripts

jherbel:
* Re-order code for better readability
* Remove global variables


racicLuka (approved):
Looks good to me

PR #251, opened 2023-07-12
Make some vagrant VM settings configurable via environment variables

jherbel:
These settings should be adjustable on a per-developer basis to account for differences in the developer machines.

CMK-13987


SoloJacobs (approved):
LGTM

PR #252, opened 2023-07-14
robotmk: remove get_commands_from_pkg

SoloJacobs:
This function makes is difficult to find the entry point of the project. Explicit is better.

CMK-14004


racicLuka (dismissed):
Looks good to me!

PR #253, opened 2023-07-14
Dev/sol/isort

SoloJacobs:


racicLuka (approved):
Looks good to me!
Just a question out of interest. You just applied isort to all the files?

PR #255, opened 2023-07-14
Add RCC as submodule and create RCC CI build job

jherbel:


jherbel:
We don't necessarily need a submodule, but it's one way of marking the RCC version we currently build (via the commit currently checked out in the submodule). We could alternatively also note this down in a file or so and then in the CI always clone the RCC repo. Would you prefer this?

SoloJacobs:
Doesn't the CI have clone a sub-module every time as well? Or are we somehow using it as an implicit caching mechanism here?

`go` includes an installation command, thus there should be a way to create the binary without `git`.

It is very hard for me to judge the approaches, without implementing a couple of them. Thus, if this is your preferred approach, then we should move forward with it.

SoloJacobs (approved):
LGTM

PR #264, opened 2023-07-20
scheduler_runner: Add function to read RCC value from config file

racicLuka:
This is supposed to read the specified configuration file and see if there is the RCC parameter defined. I assumed that it will be YES and NO, but we can easily change that depending on the bakery rule. Also, if the RCC paramater is not defined/present in the config file, I assume that the process should not be started in RCC.

As the final output it will return either True or False, meaning that the process should be started in RCC or shouldn't.

In case the file is not present or the RCC parameter is not defined or the value is not YES or NO, it exit with code 1.

### DIFF
--- a/checkmk_extensions/agent/scheduler_runner.ps1
+++ b/checkmk_extensions/agent/scheduler_runner.ps1
@@ -1,0 +1,68 @@
+function GetContentFromConfigFile {
+    [CmdletBinding()]
+
+    param (
+        [Parameter(Mandatory = $true)]
+        [string]$configFilePath
+    )
+
+    if (-not (Test-Path -Path $configFilePath -PathType Leaf)) {
+        # File doesn't exist, so we exit with message
+        Write-Error -Message "The file is not available at: $configFilePath"
+        Exit 1
+    } else {
+        # Config file is available
+        $configContent = Get-Content -Path $configFilePath
+        return $configContent
+    }
+}
+
+function DetermineRCCValueFromConfig {
+    [CmdletBinding()]
+    [OutputType([System.Boolean])]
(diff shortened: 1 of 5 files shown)


checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,68 @@
+function GetContentFromConfigFile {
+    [CmdletBinding()]
```
racicLuka, line 2: Regarding this and why it is useful:
`In PowerShell scripts, CmdletBinding() is used as a parameter attribute to define a PowerShell function or script as an advanced function. It is commonly used when creating cmdlets or more complex functions that require advanced parameter features.`

So I guess we should use it for all of our functions, since they will be somewhat complex.

jherbel (changes requested):
Let's approach this a bit differently:

- Let's a build a class `Config` which for now holds exactly one value (`rcc=true/false`). This will be easy to extend in the future.
- Furthermore, this `Config` class has a method to parse a given file. If the file doesn't exist or if the parsing fails, we raise an error an abort (no `Exit 1` if possible). The error can then be handled in some function higher up in the hierarchy.
- Let's for now assume that the config file is in JSON format, this should then be easily parsable (https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/convertfrom-json?view=powershell-7.3). Unless we have a really good reason, I would try to avoid a custom string parsing at all costs.

racicLuka:
I accidentally rebased, so it added some of Solomons changes here. I will close this pull request, and open a new one, just to keep clarity.

PR #265, opened 2023-07-21
scheduler_runner: created Config class that holds config information

racicLuka:
I need to find my way around these pull request. I got way to used to Gerrit 😅

I hope this is a step in the right direction.

This is the comment that you left on the previous PR:
```
Let's approach this a bit differently:

- Let's a build a class Config which for now holds exactly one value (tcc=true/false). This will be easy to extend in the future.
- Furthermore, this Config class has a method to parse a given file. If the file doesn't exist or if the parsing fails, we raise an error an abort (no Exit 1 if possible). The error can then be handled in some function higher up in the hierarchy.
- Let's for now assume that the config file is in JSON format, this should then be easily parsable (https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/convertfrom-json?view=powershell-7.3). Unless we have a really good reason, I would try to avoid a custom string parsing at all costs.
```


checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,63 @@
...
+            if (-Not (Test-Path -Path $configFilePath -PathType Leaf)) {
+                # Add the error message to the error collection
+                $errorCollection.Value.Add("Config file not found: $configFilePath")
+
+                # If we want to throw an ERRORRECORD object or a .NET exception, we can use the something like:
+                # $errorCollection.Value.Add((New-Object System.IO.FileNotFoundException))
+                return $null
+            }
+
+            $configFileContent = Get-Content -Raw -Path $configFilePath
```
jherbel, line 24: Will `Get-Content` throw an error if the file is missing or if the read itself fails?

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,63 @@
+# Enable StrictMode 3.0
+Set-StrictMode -Version 3.0
+
+class Config {
+    [bool]$rcc
+
+    Config() {
+        # Default constructor
```
jherbel, line 8: Is this mandatory in Powershell? Otherwise, one would of course prefer a "standard" constructor which accepts rcc as a parameter.

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,63 @@
...
+    static [Config] ParseConfigFile($configFilePath) {
+        $config = [Config]::new()
+        $errors = [System.Collections.Generic.List[string]]::new()
+
+        try {
+            $configFileContent = [Config]::GetConfigContent($configFilePath, ([ref]$errors))
+            if ($null -ne $configFileContent -and $configFileContent -ne '') {
+                $configData = ConvertFrom-Json $configFileContent
+                # Check if it contains the rcc propetry and if it's equal to YES
+                if ($configData.PSobject.Properties["rcc"] -and $configData.rcc -eq "YES") {
```
jherbel, line 44: Can't we just do
`
$config.rcc = $configData.PSobject.Properties["rcc"]
`
Or, to be safe (not sure if this is proper Powershell syntax)
`
$config.rcc = $configData.PSobject.Properties["rcc"] -as [bool]
`

jherbel (changes requested):

checkmk_extensions/agent/scheduler_runner.ps1
jherbel: Ok, so then we don't have to check ourselves if the file exists, right? Get-Content will tell us if it doesn't.

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: I was only able to make this work, if in the json file the value for `rcc` is either `true` or `false`.

In all other cases, for example if we use `Yes` and `No`, it will evaluate the existance of a string as `true`, so it will lead to false positives.

Since we can influence the bakery rule, we can make sure that the value in the json will be a `bool`.

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: No, it's not. We can avoid the default constructor and Powershell will on it's own assign the value `$false` to the rcc parameter.

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: Yes, we can make sure that it raises it. I will just add a -ErrorAction flag to it.

checkmk_extensions/agent/scheduler_runner.ps1
jherbel: That's fine, let's use `-as [bool]` (similar to what pydantic does) and if that leads to a false positive, it's definitely not our fault

jherbel (changes requested):
Looks already quite good :) Last point: Do we actually need to collect any errors? Can't we just catch and throw it again right away? Or does `throw` not terminate the current function?

racicLuka:
> 

Actually, you are right. We don't need to collect them, we can just `throw` them right away. I adapted the code.

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,36 @@
...
+            $configFileContent = [Config]::GetConfigContent($configFilePath)
+            if ($null -ne $configFileContent -and $configFileContent -ne '') {
+                $configData = $configFileContent | ConvertFrom-Json -ErrorAction Stop
+
+                # Assign the value from the config file to rcc
+                $config.rcc = $configData.PSobject.Properties["rcc"].Value -as [bool]
+                }
+        }
+        catch {
+            Write-Host "An error occurred while parsing the config file" -ForegroundColor Red
```
jherbel, line 31: No need for any custom message, we can directly simply do `throw`

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,36 @@
+# Enable StrictMode 3.0
+Set-StrictMode -Version 3.0
+
+class Config {
+    [bool]$rcc
+
+    static [string] GetConfigContent($configFilePath) {
+        try {
+            $configFileContent = Get-Content -Raw -Path $configFilePath -ErrorAction Stop
```
jherbel, line 9: What would happen if we simply wrote
`
return Get-Content -Raw -Path $configFilePath -ErrorAction Stop
`
Would this not throw the error?

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+
+    # Method to parse the config file (JSON) and set the value of rcc
+    static [Config] ParseConfigFile($configFilePath) {
+        $config = [Config]::new()
+
+        try {
+            $configFileContent = [Config]::GetConfigContent($configFilePath)
+            if ($null -ne $configFileContent -and $configFileContent -ne '') {
```
jherbel, line 23: Can we even encounter `$null` here? Also, I don't think we have to check for empty strings ourselves. Json will tell us if there is a problem.

jherbel (changes requested):

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: We could encounted `$null` if, for example the json file is present, but empty. The solution would be the same as above: to make the `rcc` parameter have a default value via the bakery rule.

You are right, we don't need to check for the empty string.

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,29 @@
+# Enable StrictMode 3.0
+Set-StrictMode -Version 3.0
+
+class Config {
+    [bool]$rcc
+
+    # Constructor
+    # This is a way to make the rcc property mandatory
+    Config([bool]$rcc) {
```
jherbel, line 9: Nice :+1:

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -0,0 +1,29 @@
...
+        $this.rcc = $rcc
+    }
+
+    # Method to parse the config file (JSON) and set the value of rcc
+    static [Config] ParseConfigFile($configFilePath) {
+
+        $configFileContent = Get-Content -Raw -Path $configFilePath -ErrorAction Stop
+
+        try {
+            $configData = $configFileContent | ConvertFrom-Json -ErrorAction Stop
```
jherbel, line 19: Could this be moved outside the try-catch-block?

jherbel (approved):
In this case, you can squash all commits, I think GitHub offers this option upon merging

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka, line 19: I needed to leave this in the try-catch block because even if the `ConvertFrom-Json` function has an exception the script would continue running. This was the only way, I can think of, to ensure that it will stop with the execution.

PR #267, opened 2023-07-21
Dev/sol/add some coverage

SoloJacobs:


SoloJacobs:
Nevermind. This does not work as expected.

PR #268, opened 2023-07-24
Install PSScriptAnalyzer in setup

jherbel:


SoloJacobs (commented):
Based on this post, I think we should omit the -Force:
https://stackoverflow.com/questions/34749333/powershell-guidelines-for-confirm-force-and-whatif
From what I understand, we don't really want it overwrite existing things... Or is this necessary for the Github actions?

jherbel:
@SoloJacobs I removed -Force

SoloJacobs (approved):
LGTM!

PR #271, opened 2023-07-24
Add static analysis for v2 directory

SoloJacobs:
* mypy
* pylint
* black
* isort

pytest is missing, and so are any tests. `runner.py` was moved, such that mypy knows not to check the directory `v2/data/`.


racicLuka (dismissed):
Looks good to me!

PR #274, opened 2023-07-26
Retryer: Implement incremental retries

jherbel:
CMK-14090

### DIFF
--- a/v2/src/runner.py
+++ b/v2/src/runner.py
@@ -1,19 +1,30 @@
 """RetryStrategy. """
 
 import dataclasses
+import enum
 import pathlib
 import uuid
 from collections.abc import Sequence
 
 
+class _RetryStrategy(enum.Enum):
+    INCREMENTAL = "incremental"
+    COMPLETE = "complete"
+
+
 @dataclasses.dataclass(frozen=True)
-class _RunnerSpec:
+class _RunnerSpec:  # pylint: disable=too-many-instance-attributes
     python_executable: pathlib.Path
     robot_target: pathlib.Path
     outputdir: pathlib.Path
-    output: str
+    output_name: str
+    previous_output: pathlib.Path | None
     variablefile: pathlib.Path | None
     argumentfile: pathlib.Path | None
+    retry_strategy: _RetryStrategy
+
+    def output(self) -> pathlib.Path:
+        return self.outputdir / f"{self.output_name}.xml"
 
     def _check(self) -> bool:
         paths_to_check = [
@@ -22,6 +33,7 @@ class _RunnerSpec:
             self.outputdir,
             self.variablefile,
             self.argumentfile,
+            self.previous_output,
         ]
         return any(not path.exists() for path in paths_to_check if path is not None)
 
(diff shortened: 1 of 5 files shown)


v2/src/runner.py
```diff
@@ -64,11 +75,12 @@ def _create_commands(spec: _RetrySpec) -> Sequence[str]:
             python_executable=spec.python_executable,
             robot_target=spec.robot_target,
             outputdir=spec.outputdir(),
-            output=str(
-                i
-            ),  # Ensure the `robot` command does not overwrite previous runs
+            # Ensure the `robot` command does not overwrite previous runs
+            output=spec.outputdir() / f"{i}.xml",
```
jherbel, line 79: Since we have to pass an absolute path for `previous_output` (robot framework does _not_ prepend `outputdir` if `previous_output` is relative), it seemed to me that it would make sense to use an absolute path here as well, st. these two are for sure in sync.

v2/tests/test_runner.py
```diff
@@ -3,18 +3,40 @@
...
     spec = runner._RunnerSpec(  # pylint: disable=protected-access
         python_executable=pathlib.Path("/usr/bin/python3"),
         robot_target=pathlib.Path("~/suite/calculator.robot"),
         outputdir=pathlib.Path("/tmp/outputdir/"),
-        output="output",
+        output=pathlib.Path("/tmp/outputdir/output.xml"),
+        previous_output=None,
         variablefile=None,
         argumentfile=None,
+        retry_strategy=runner._RetryStrategy.COMPLETE,  # pylint: disable=protected-access
```
SoloJacobs, line 15: I don't think the _RetryStrategy belongs here, I would rather call this parameter rerun-failed, with rerun failed having time `Output` | None. The `Retrier` is responsible for building `Runners`. But since we are prototyping this is not a hard requirement. I want to do some cleaning up anyway.

v2/src/runner.py
SoloJacobs: That makes sense... Although I don't like specifying duplicate information inside the struct. Maybe having a helper function, which exposes the output as an absolute path instead?

v2/src/runner.py
```diff
@@ -1,19 +1,27 @@
...
     robot_target: pathlib.Path
     outputdir: pathlib.Path
-    output: str
+    output: pathlib.Path
+    previous_output: pathlib.Path | None
     variablefile: pathlib.Path | None
     argumentfile: pathlib.Path | None
+    retry_strategy: _RetryStrategy
 
     def _check(self) -> bool:
```
SoloJacobs, line 26: Should probably be updated...

SoloJacobs (approved):
I have added some comments, which would warrant discussion. These comments are about creating good abstractions rather than getting us closer to a running solution... Thus, I think we can ignore them, I think we should just merge this.

v2/src/runner.py
jherbel: Done.

v2/src/runner.py
jherbel: Done

v2/tests/test_runner.py
jherbel: I agree, not ideal, can be improved later.

PR #275, opened 2023-07-26
Retryer: Implement merge command creation

jherbel:
CMK-14116

### DIFF
--- a/v2/src/runner.py
+++ b/v2/src/runner.py
@@ -4,7 +4,7 @@ import dataclasses
 import enum
 import pathlib
 import uuid
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
 
 
 class _RetryStrategy(enum.Enum):
@@ -25,6 +25,20 @@ class _RunnerSpec:  # pylint: disable=too-many-instance-attributes
 
     def output(self) -> pathlib.Path:
         return self.outputdir / f"{self.output_name}.xml"
+
+    def command(self) -> str:
+        robot_command = f"{self.python_executable} -m robot "
+        if self.variablefile is not None:
+            robot_command += f"--variablefile={self.variablefile} "
+        if self.argumentfile is not None:
+            robot_command += f"--argumentfile={self.argumentfile} "
+        if self.retry_strategy is _RetryStrategy.INCREMENTAL and self.previous_output:
+            robot_command += f"--rerunfailed={self.previous_output} "
+        return robot_command + (
+            f"--outputdir={self.outputdir} "
+            f"--output={self.output()} "
+            f"{self.robot_target}"
+        )
 
     def _check(self) -> bool:
         paths_to_check = [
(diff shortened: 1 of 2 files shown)


v2/src/runner.py
```diff
@@ -72,9 +72,28 @@ def _create_command(spec: _RunnerSpec) -> str:
...
+    final_output: pathlib.Path,
+) -> str:
+    return (
+        f"{python_executable} -m robot.rebot --output={final_output} --report=NONE --log=NONE "
+        f"{' '.join(str(variant_output) for variant_output in variants_outputs)}"
+    )
+
+
+@dataclasses.dataclass(frozen=True)
+class _RetryCommands:
```
jherbel, line 88: I am not really happy with this name and thus open to suggestions

v2/src/runner.py
```diff
@@ -83,11 +102,19 @@ def _create_commands(spec: _RetrySpec) -> Sequence[str]:
...
-    return commands
+        variants.append(_create_variant_command(runner_cfg))
+        outputs.append(runner_cfg.output())
+
+    return _RetryCommands(
+        variants=variants,
+        merge=_create_merge_command(
+            python_executable=spec.python_executable,
+            variants_outputs=outputs,
+            final_output=spec.outputdir() / "merged.xml",
```
jherbel, line 118: I simply hard-coded this for now. We should probably make this configurable or at least store the name in a variable once it's clear how we post-process these files further.

v2/tests/test_runner.py
```diff
@@ -1,42 +1,104 @@
...
+            (
+                "/usr/bin/python3 -m robot --variablefile=~/suite/retry.yaml "
+                "--rerunfailed=/tmp/outputdir/383783f41d0243b19d6f205f4d492d95/0.xml "
+                "--outputdir=/tmp/outputdir/383783f41d0243b19d6f205f4d492d95 "
+                "--output=/tmp/outputdir/383783f41d0243b19d6f205f4d492d95/1.xml "
+                "~/suite/calculator.robot"
+            ),
+        ],
+        merge=(
+            "/usr/bin/python3 -m robot.rebot "
```
SoloJacobs, line 98: Happen happens if we have RetryStrategy, which runs twice, but has three variants specified? I believe the interface of the merge command is the result of command executions, rather than of the strategy?

v2/src/runner.py
SoloJacobs: It seems to me describe the content well enough. So, I would rethink the class but see my comment about the merge command.

SoloJacobs (changes requested):
My gut feeling is that this that our implementation is straying a bit of target here.

v2/src/runner.py
```diff
@@ -57,37 +58,75 @@ def outputdir(self) -> pathlib.Path:
...
+    @property
+    def command(self) -> str:
+        ...
+
+
+@dataclasses.dataclass(frozen=True)
+class _AttemptImpl:
+    runner_spec: _RunnerSpec
+
+    @property
```
SoloJacobs, line 75: Oti: Just for the purpose of curiosity: What is the benefit of the @property decorator here? I personally like my function calls to look like function calls.

SoloJacobs (dismissed):
LGTM

SoloJacobs (approved):
LGTM

PR #280, opened 2023-07-31
Dev/sol/rccer

SoloJacobs:


SoloJacobs:
This commit https://github.com/elabit/robotmk/tree/dev/sol/add_examples
was relatively instrumental as a smoke test, but it is not really a clean unit test. Thus, I didn't include it here.

jherbel (approved):

PR #281, opened 2023-07-31
scheduler_runner: Create command

racicLuka:
Create command depending on whether we should use rcc or not.

I am still not sure what the "arguments" here need to be. Would be nice if we can take a look together. I tried to conclude it from Simon's ps1 script, but I am not sure I understand it fully.


checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -27,3 +27,19 @@ class Config {
...
+           $errorMessage = "Error: 'rcc.exe' binary not found in the current folder."
+            throw New-Object System.IO.FileNotFoundException -ArgumentList $errorMessage
+        }
+        # TODO: Add the arguments.
+        # What do I need?
+        # Simon is calculating the blueprint here
+        return ("./rcc.exe")
+    }
+    else {
+        $pythonExe = (Get-Command python).Source # Should we use python or python3 here?
```
jherbel, line 42: lets use python3

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -27,3 +27,19 @@ class Config {
...
+        return ($pythonExe)
```
jherbel, line 43: We want to call something
`python3 -m robotmk.scheduler --config=../config/robotmk.json`
The exact details (config file name and path) remain to be seen, they can be easily adapted later

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -27,3 +27,19 @@ class Config {
...
+function CreateCommand([Config]$config) {
+    if ($config.rcc) {
+        if (-Not (Test-Path -Path "./rcc.exe" -PathType Leaf)) {
+           $errorMessage = "Error: 'rcc.exe' binary not found in the current folder."
+            throw New-Object System.IO.FileNotFoundException -ArgumentList $errorMessage
+        }
+        # TODO: Add the arguments.
+        # What do I need?
+        # Simon is calculating the blueprint here
+        return ("./rcc.exe")
```
jherbel, line 39: Here, we can assume that we have a robot.yaml (let's say also in ../config) with a single task, so we want to execute something like
`rcc run --robot ../config`
This can be easily adapter later on.

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -27,3 +27,19 @@ class Config {
         return [Config]::new($use_rcc)
     }
 }
+function CreateCommand([Config]$config) {
+    if ($config.rcc) {
+        if (-Not (Test-Path -Path "./rcc.exe" -PathType Leaf)) {
```
jherbel, line 32: No need to check this here. We simply try to execute later on.

jherbel (changes requested):
Whenever I wrote ../config, we should use absolute paths if possible, where by we compute ../config wrt. to the location of the script

checkmk_extensions/agent/scheduler_runner.ps1
simonmeggle: On Windows you typically do not need the version suffix => `python` is enough

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -27,3 +27,25 @@ class Config {
         return [Config]::new($use_rcc)
     }
 }
+function CreateCommand([Config]$config) {
```
jherbel, line 30: Return type (if possible)?

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -27,3 +27,25 @@ class Config {
...
+        $pythonExe = (Get-Command python).Source
+        $moduleName = "robotmk.scheduler"
+        $configPath = Join-Path $parentDir "config\robotmk.json"
+
+        $argumentsList = "-m $moduleName --config $configPath"
+    }
+
+    return [PSCustomObject]@{
+        Executable = if ($config.rcc) { $rccPath } else { $pythonExe }
+        Arguments = $argumentsList
```
jherbel, line 49: Is this now an actual list (sorry, my PowerShell syntax knowledge is very poor ..)? Basically, it has to work as input for
[-ArgumentList] <String[]> in Start-Process, right?

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -27,3 +27,25 @@ class Config {
...
+    else {
+        $pythonExe = (Get-Command python).Source
+        $moduleName = "robotmk.scheduler"
+        $configPath = Join-Path $parentDir "config\robotmk.json"
+
+        $argumentsList = "-m $moduleName --config $configPath"
+    }
+
+    return [PSCustomObject]@{
+        Executable = if ($config.rcc) { $rccPath } else { $pythonExe }
```
jherbel, line 48: We can just define a variable `$executable` above (which is either `Join-Path $PSScriptRoot "rcc.exe"` or `(Get-Command python).Source`), then we don't have to check any condition here.

jherbel (changes requested):

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: I have a commit ready that will make a list out of this, because at this point it's just a long string. But, Start-Process also works with a string as well.

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: The return type can't be directly defined, but it is "hinted" at the bottom with  `return [PSCustomObject]`.
I can make it more explicit and create a class `Command` that would have the executable and argumentsList and then return it. Does that sound OK?

checkmk_extensions/agent/scheduler_runner.ps1
jherbel: Ok, then it can also stay as is

checkmk_extensions/agent/scheduler_runner.ps1
jherbel: That's a good idea, let's do that :+1:

jherbel (approved):

PR #291, opened 2023-08-02
scheduler_runner: StartSchedulerRunner

racicLuka:
I messed up something in the last PR, so I closed it. Sorry for that!

### DIFF
--- a/checkmk_extensions/agent/scheduler_runner.ps1
+++ b/checkmk_extensions/agent/scheduler_runner.ps1
@@ -57,7 +57,13 @@ class Command {
     }
 }
 
-function StartSchedulerRunner([string]$configFilePath) {
+function StartSchedulerRunner {
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory=$true)]
+        [string]$configFilePath
+    )
+
     $configFileContent = [Config]::ParseConfigFile($configFilePath)
 
     $command = [Command]::CreateCommand($configFileContent)
--- a/checkmk_extensions/agent/scheduler_runner.ps1
+++ b/checkmk_extensions/agent/scheduler_runner.ps1
@@ -64,9 +64,9 @@ function StartSchedulerRunner {
         [string]$configFilePath
     )
 
-    $configFileContent = [Config]::ParseConfigFile($configFilePath)
+    $config = [Config]::ParseConfigFile($configFilePath)
 
-    $command = [Command]::CreateCommand($configFileContent)
+    $command = [Command]::CreateCommand($config)
     $parentDir = Split-Path $PSScriptRoot -Parent
     $logPath = Join-Path $parentDir "log\robotmk\scheduler_runner.log"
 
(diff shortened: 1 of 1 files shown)


checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -56,3 +56,87 @@ class Command {
...
 }
+
+function StartSchedulerRunner {
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory=$true)]
+        [string]$configFilePath
+    )
+
+    $configFileContent = [Config]::ParseConfigFile($configFilePath)
```
jherbel, line 67: `config = ...`
We get an actual `Config` object here, right?

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -56,3 +56,87 @@ class Command {
...
+    # $processInfo.WorkingDirectory -> Using this we can configure from where the Process will be started
+    $processInfo.WorkingDirectory = $PSScriptRoot
+
+    # Create process and assign processInfo to it
+    $process = New-Object System.Diagnostics.Process
+    $process.StartInfo = $processInfo
+
+
+    try {
+        $process.Start() | Out-Null
```
jherbel, line 90: This should happen inside the loop. If the process crashes, we want to re-start it (should never happen, of course).

racicLuka:
Regarding the logging. I am looking into some other options for it.

One is `PSLogging` - But I don't particularly like it, because it needs to be installed and imported, but doesn't have much advantages.

The other one is `Start-Transcript` - This is a built-in command in Powershell. It looks like it can be useful, because of some nice features that it has, but I need to familiarize myself with it more, because it isn't as straight forward as it look at first glance.

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -56,3 +56,87 @@ class Command {
...
+        }
+    }
+    catch {
+        # TODO: Handle errors
+        WriteLog -Message $_.Exception.Message -LogPath $logPath
+        throw $_.Exception.Message
+    }
+}
+
+function WriteLog {
```
jherbel, line 114: I think we will likely want to do more than logging (report the error to the Checkmk server in the form of a section), but this can stay as is for now. We can extend it in a follow-up step.

jherbel (changes requested):

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: I don't know what exactly you are reffering to.

I created it so that we just need the path to the config file as an input to start the scheduler-runner, so we expect a string. Then we can just start a process with `StartSchedulerRunner($pathToConfig)`.

jherbel: I was referring to the variable name `$configFileContent`. `[Config]::ParseConfigFile($configFilePath)` returns a `Config` object, right? So `$config` seems to be a more fitting name. `$configFileContent` would suggest that we get the raw, unparsed file content.

jherbel (approved):

PR #295, opened 2023-08-09
Add editor dependencies

SoloJacobs:
For obscure reasons, python-lsp-server needs to be installed into the virtual environment.

Updating the lock in a previous commit was not necessary, since the lock was up-to-date.


racicLuka (approved):
LGTM

PR #299, opened 2023-08-23
Run RCC commands as controller 'robotmk' and in dedicated suite space

jherbel:


v2/robotmk/environment.py
```diff
@@ -31,6 +34,7 @@ def wrap_for_execution(self, command: Sequence[str]) -> list[str]:
             str(self.binary),
             "task",
             "script",
+            *self._controller_and_space_args(),
```
SoloJacobs, line 37: OTI: was under the impression that the `-r` argument would automatically select the correct space.

jherbel: I don't think that `-r` has any impact on the space. From my understanding, unless explicitly specified, you always end up with the default (`"user"`).

PR #301, opened 2023-08-23
Add session layer

jherbel:
For now, we only implement running in the current session (can be either headed or headless). In the future, we will add the option to run in the session of another user.

CMK-14088


SoloJacobs (approved):
LGTM

PR #302, opened 2023-08-23
Move logic from `runner._RunnerSpec` to `runner.Attempt`

jherbel:
Implementing running in a different session via tasks will require access to the output directory, the id and the index of an attempt. Once `runner.Attempt` has these fields, we might as well move the logic to build a robot command there.

CMK-14088


SoloJacobs (approved):
LGTM

PR #305, opened 2023-08-23
Scheduler runner: Rename `Command` -> `SchedulerExecutionCommand`

jherbel:
The name "Command" causes problems because of PSAvoidUsingCmdletAliases: 'Command' is implicitly aliasing 'Get-Command' because it is missing the 'Get-' prefix. This can introduce possible problems and make scripts hard to maintain. Please consider changing command to its full name.

CMK-14327


racicLuka (commented):
LGTM

racicLuka (approved):

PR #307, opened 2023-08-24
Scheduler runner: Fix execution logic

jherbel:
Before, the scheduler runner continuously started scheduler processes, however, we want to run only one scheduler process at a time. Logging and error handling might still need improvement, but at least this change makes the scheduler runner usable.

CMK-14327


racicLuka (approved):
LGTM

PR #309, opened 2023-08-25
Scheduler runner: Implement full configuration

jherbel:
Implement parsind and usage of configuration object with all necessary fields. The configuration will be shared with the scheduler and the agent plugin that sends the final output to Checkmk.


racicLuka (approved):
Looks really good

jherbel:
Hi @racicLuka , sry, it wasn't fully ready yet before

racicLuka (approved):

PR #312, opened 2023-08-25
Fix check-all mode in ci script

jherbel:
The exit code of check-all was the exit code of the last command executed in this mode, which is obviously wrong. We now return the sum of the individual exit codes instead.


SoloJacobs (approved):
LGTM

PR #313, opened 2023-08-25
Scheduler: Switch to joint config format

jherbel:
The configuration will be shared with the scheduler runner and the agent plugin which produces the final section.

* RCC usage is configurable globally, not on a per-suite bases.
* Working directory is configurable globally, not on a per-suite bases.

CMK-14346

### DIFF
--- a/v2/robotmk/scheduler.py
+++ b/v2/robotmk/scheduler.py
@@ -1,16 +1,17 @@
 """Scheduler"""
 
 import argparse
+import dataclasses
 import datetime
 import pathlib
 import subprocess
-from collections.abc import Iterable, Mapping, Sequence
-from typing import Final
+from collections.abc import Iterable, Iterator, Mapping, Sequence
+from typing import Final, Literal
 from uuid import uuid4
 
 from apscheduler.schedulers.blocking import BlockingScheduler  # type: ignore[import]
 from apscheduler.triggers.interval import IntervalTrigger  # type: ignore[import]
-from pydantic import BaseModel
+from pydantic import BaseModel, TypeAdapter
 from robot import rebot  # type: ignore[import]
 
 from robotmk.environment import RCCEnvironment, ResultCode, RobotEnvironment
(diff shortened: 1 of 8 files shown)


v2/robotmk/scheduler.py
```diff
@@ -1,96 +1,151 @@
 """Scheduler"""
```
SoloJacobs, line 1: Should this update the `windows.json` files also?

v2/robotmk/scheduler.py
```diff
@@ -1,96 +1,151 @@
...
     return scheduler
 
 
+@dataclasses.dataclass(frozen=True)
+class _RCCEnvironmentSpec:
+    binary_path: pathlib.Path
+    robot_yaml_path: pathlib.Path
+
+
+def _rcc_env_specs(
```
SoloJacobs, line 84: Probably more appropriate as class method

SoloJacobs (approved):
LGTM but I have some comments.

v2/robotmk/scheduler.py
jherbel: I reworked this, I think it's formulated in a better way now. Good suggestion, thanks!

v2/robotmk/scheduler.py
```diff
@@ -1,96 +1,151 @@
...
+
 
+class _RCCSuiteConfig(_SuiteConfig, frozen=True):
+    robot_yaml_path: pathlib.Path
 
-class _SchedulerConfig(BaseModel, frozen=True):
-    suites: Mapping[str, _SuiteConfig]
 
+class _ConfigSystemPython(BaseModel, frozen=True):
+    rcc: None
```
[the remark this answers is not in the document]
jherbel: In the original version, it couldn't be removed, otherwise, any config would deserialize to the system python case. However, I made this more explicit now anyway.

SoloJacobs (approved):
LGTM

PR #314, opened 2023-08-28
Scheduler: Write output of suite retries to `[scheduler_working_directory]/[suite_name]/[timestamp]`

jherbel:
Also name RF HTML logs `[retry_index].html`.

### DIFF
--- a/v2/robotmk/runner.py
+++ b/v2/robotmk/runner.py
@@ -48,8 +48,8 @@ class Attempt:
     argument_file: pathlib.Path | None
     retry_strategy: RetryStrategy
 
-    def output_file(self) -> pathlib.Path:
-        return self._output_file(self.index)
+    def output_xml_file(self) -> pathlib.Path:
+        return self.output_directory.joinpath(f"{self.index}.xml")
 
     def command(self) -> list[str]:
         robot_command = [str(PYTHON_EXECUTABLE), "-m", "robot"]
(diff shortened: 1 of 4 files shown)


v2/robotmk/runner.py
```diff
@@ -58,15 +58,18 @@ def command(self) -> list[str]:
...
             f"--outputdir={self.output_directory}",
-            f"--output={self.output_file()}",
+            f"--output={self.output_xml_file()}",
+            f"--log={self._output_file(self.index, 'html')}",
             str(self.robot_target),
         ]
 
-    def _output_file(self, index: int) -> pathlib.Path:
-        return self.output_directory.joinpath(f"{index}.xml")
+    def _output_file(self, index: int, suffix: Literal["xml", "html"]) -> pathlib.Path:
```
SoloJacobs, line 71: This seems like one indirection more than needed, and the signature is mixing `self.index` and `index`. Make this a classmethod? But I prefer removing the method.

v2/robotmk/scheduler.py
```diff
@@ -148,7 +154,10 @@ def __init__(self, suite_specification: _SuiteSpecification) -> None:
 
     def __call__(self) -> None:
         retry_spec = RetrySpec(
-            id_=uuid4(),
+            identifier=Identifier(
+                name=self._suite_spec.name,
+                timestamp=datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S"),
```
SoloJacobs, line 159: Could we get clashes here, if somebody updates there local time? Should we maybe fallback to utc time here?

OTI: I think just using ISO8601 would be fine here: https://stackoverflow.com/questions/2150739/iso-time-iso-8601-in-python

v2/robotmk/runner.py
```diff
@@ -3,13 +3,18 @@
...
 from collections.abc import Iterator, Sequence
 from typing import Final
 
 PYTHON_EXECUTABLE: Final = pathlib.Path("python")
 
 
+@dataclasses.dataclass(frozen=True)
+class Identifier:
+    name: str
+    timestamp: str
```
SoloJacobs, line 15: OTI: int / str / datetime? so many options

jherbel: Let's go with str for now, since we use this in multiple places. Otherwise, we would have either duplicate code or some sort of helper function to make this a str.

v2/robotmk/runner.py
jherbel: Done

v2/robotmk/scheduler.py
jherbel: I actually first tried the ISO format, but ":" is not allowed in Windows directory names. I now use the ISO format with ":" replaced by ".".

SoloJacobs: Ah, now that I understand the reasoning, I prefer the previous version :smile: But the important part was the timezone anyway. Let's not iterate further on this version, they don't seem important at the moment.

SoloJacobs (approved):
LGTM

PR #315, opened 2023-08-29
Scheduler: Write final JSON output files

jherbel:
For now, these files contain the XML data only.


v2/robotmk/scheduler.py
```diff
@@ -197,6 +201,29 @@ def _run_attempts_until_successful(
             break
         return outputs
 
+    def _write_result_file(self, merged_xml_path: pathlib.Path) -> None:
```
SoloJacobs, line 204: This is fine for me as a first version. However, I thought decided that moving files into the `results_directory` should be atomic?

v2/robotmk/scheduler.py
```diff
@@ -173,9 +177,9 @@ def __call__(self) -> None:
             return  # Untested
 
         final_output = retry_spec.output_directory() / "merged.xml"
-
         rebot(*outputs, output=final_output, report=None, log=None)
-        self._final_outputs.append(final_output)
+
+        self._write_result_file(final_output)
```
SoloJacobs, line 182: OTI: This function now has multiple side-effects. I think we should maybe review this with Mathias?

SoloJacobs (approved):
Some otis

v2/robotmk/scheduler.py
jherbel: I had forgotten about that, you are right. Thanks!

v2/robotmk/scheduler.py
jherbel: I would say that after this commit, the scheduler is in a minimum viable product stage. I would suggest to do two steps:

1. We sit together internally and discuss the code quality. We can also do immediate improvements such as renamings.
2. We discuss/review our code with eg. Mathias.

What do you think about this?

SoloJacobs (approved):

PR #317, opened 2023-08-30
Add agent plugin

SoloJacobs:
CMK-14376


jherbel:
As such it of course makes sense to use shared code regarding the configuration. However, I am not sure if we can do this in production. When being deployed by the bakery, the scheduler runner and the agent plugin won't be located in the same folder. Let's discuss this tomorrow.

SoloJacobs:
Updated to simple logic duplication inside the plugin.

jherbel (approved):

PR #322, opened 2023-09-04
scheduler: extract public api into module

SoloJacobs:
CMK-14423

### DIFF
--- a/v2/robotmk/api.py
+++ b/v2/robotmk/api.py
@@ -1,0 +1,5 @@
+from pydantic import BaseModel
+
+
+class Result(BaseModel, frozen=True):
+    xml: str
--- a/v2/robotmk/api.py
+++ b/v2/robotmk/api.py
@@ -1,6 +1,36 @@
 from pydantic import BaseModel
 
+from robotmk import parse_xml
+
+
+class Test(BaseModel, frozen=True):
+    name: str
+    id_: str
+
 
 class Result(BaseModel, frozen=True):
     suite_name: str
+    tests: list[Test]
     xml: str
+
+
+# TODO: This function depends on `parse_xml` and therefore   # pylint: disable=fixme
+# cannot be part of the API! We have to find a different location.
+def create_result(suite_name: str, xml: str) -> Result:
+    rebot = parse_xml.parse_rebot(xml)
+    tests = _obtain_tests(rebot)
+    return Result(
+        suite_name=suite_name,
+        tests=[Test(name=t.name, id_=t.id_) for t in tests],
+        xml=xml,
+    )
+
+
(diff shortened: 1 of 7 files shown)


v2/tests/test_parse_xml.py
```diff
@@ -0,0 +1,7 @@
+from robotmk import parse_xml
+
+
+def test_parse_rebot() -> None:
+    with open("v2/tests/rebot.xml", "r", encoding="utf-8") as file:
```
jherbel, line 5: Does this also work if I execute only this test?

v2/robotmk/api.py
```diff
@@ -0,0 +1,34 @@
...
+    id_: str
+
+
+class Result(BaseModel, frozen=True):
+    suite_name: str
+    tests: list[Test]
+    xml: str
+
+
+def create(suite_name: str, xml: str) -> Result:
```
jherbel, line 17: `create_result` maybe? Otherwise, this could also become a classmethod of `Result`.

v2/robotmk/parse_xml.py
```diff
@@ -0,0 +1,69 @@
+import enum
```
jherbel, line 1: Can you briefly explain why we need this? I thought that RF offers parsing functionality for the XML data. Is the functionality offered by RF insufficient for us?

jherbel (changes requested):

v2/tests/test_parse_xml.py
SoloJacobs: These work no problem:
```
pytest v2/tests/test_parse_xml.py
pytest v2/tests/test_parse_xml.py::test_parse_rebot
```
Or did you mean something else?

v2/robotmk/api.py
SoloJacobs: I did not make this a `classmethod`, because `create` depends on `parse_xml`. `parse_xml` does not belong into the `robotmk_api` and thus `create` cannot be part of the `robotmk_api`. In fact, that function does not belong in that module, but into it's own module. However, until we actually start addressing the API problem, these considerations should not really affect our product, so I didn't create the extra module.

`create_result` will probably have to be changed as well, when we add a second JSON model, but for now this is ok with me.

v2/tests/test_parse_xml.py
jherbel: No, this is what I meant.

v2/robotmk/api.py
jherbel: Ok.

v2/robotmk/parse_xml.py
jherbel: :+1:

jherbel (approved):

PR #323, opened 2023-09-04
Added some error handling and logging

racicLuka:
It will still need some work, but for now it's logging the errors in both the log file and creating a agent output-like file.

### DIFF
--- a/checkmk_extensions/agent/scheduler_runner.ps1
+++ b/checkmk_extensions/agent/scheduler_runner.ps1
@@ -43,13 +43,8 @@ class Config {
     }
 
     static [Config] ParseConfigFile([string]$Path) {
-        try {
-            $configFileContent = Get-Content -Raw -Path $Path
-            $configData = ConvertFrom-Json -InputObject $configFileContent
-        }
-        catch {
-            throw
-        }
+        $configFileContent = Get-Content -Raw -Path $Path
+        $configData = ConvertFrom-Json -InputObject $configFileContent
 
         $resultsDir = $configData.results_directory -as [string]
         $logDir = $configData.log_directory -as [string]
@@ -121,18 +116,10 @@ function StartSchedulerRunner {
         [string]$ConfigFilePath
     )
 
-    try {
-        $config = [Config]::ParseConfigFile($ConfigFilePath)
-    }
-    catch {
-        # What should the path of the log files be?
-        # Here we still don't have access to the defined path from the config file
-        WriteLogAndException -Message $_.Exception.Message -LogPath "" -ExceptionLogPath ""
-        throw
-    }
+    $config = [Config]::ParseConfigFile($ConfigFilePath)
 
     $selfLogPath = Join-Path $config.LogDirectory "scheduler_runner.log"
-    $exceptionLogPath = Join-Path $config.LogDirectory "scheduler_runner"
+    $exceptionLogPath = Join-Path $config.ResultsDirectory "scheduler_runner"
     $schedulerStdoutLogPath = Join-Path $config.LogDirectory "scheduler_stdout.log"
     $schedulerStderrLogPath = Join-Path $config.LogDirectory "scheduler_stderr.log"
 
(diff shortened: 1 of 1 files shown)


checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -116,17 +121,36 @@ function StartSchedulerRunner {
...
+    }
+    catch {
+        # What should the path of the log files be?
+        # Here we still don't have access to the defined path from the config file
+        WriteLogAndException -Message $_.Exception.Message -LogPath "" -ExceptionLogPath ""
+        throw
     }
 
     $selfLogPath = Join-Path $config.LogDirectory "scheduler_runner.log"
+    $exceptionLogPath = Join-Path $config.LogDirectory "scheduler_runner"
```
racicLuka, line 135: Here we are defining the path of the agent output-like log file, but I am not sure what it should be, so I just chose this.

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -43,8 +43,13 @@ class Config {
     }
 
     static [Config] ParseConfigFile([string]$Path) {
-        $configFileContent = Get-Content -Raw -Path $Path
-        $configData = ConvertFrom-Json -InputObject $configFileContent
+        try {
```
jherbel, line 46: I don't think we need that. The line `$ErrorActionPreference = "Stop"` should basically take care of this. I thought I had tested this. Is this not the case?

checkmk_extensions/agent/scheduler_runner.ps1
```diff
@@ -116,17 +121,36 @@ function StartSchedulerRunner {
...
-
-    if (-not (Test-Path $config.LogDirectory)) {
-        New-Item -Path $config.LogDirectory -ItemType Directory
+    try {
+        $config = [Config]::ParseConfigFile($ConfigFilePath)
+    }
+    catch {
+        # What should the path of the log files be?
+        # Here we still don't have access to the defined path from the config file
+        WriteLogAndException -Message $_.Exception.Message -LogPath "" -ExceptionLogPath ""
```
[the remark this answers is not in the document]
jherbel: If we need to log this error should be part of the specifications provided by Simon (as well as where to log it).

checkmk_extensions/agent/scheduler_runner.ps1
jherbel: Not sure what you mean by "agent output-like log file". The config object has a `ResultsDirectory` field. In this directory, we can write agent output.

jherbel (changes requested):

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: Makes sense. I will adjust it.

checkmk_extensions/agent/scheduler_runner.ps1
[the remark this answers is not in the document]
racicLuka: In the specification it says that we should log all the occuring errors, including this one.

checkmk_extensions/agent/scheduler_runner.ps1
racicLuka: It will cause the script to stop executing, but we will not log the error. Simon wrote in the specification that we should log it, so this is the way to ensure that it will be logged.

racicLuka: No need for change, since we will not log this error.

checkmk_extensions/agent/scheduler_runner.ps1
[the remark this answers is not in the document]
racicLuka: We decided to not log this error.

jherbel (changes requested):
There seems to be some confusion regarding `$exceptionLogPath`. The results directory is not intended for logging purposes. There, we write JSON-serialized result files which are then transported to the Checkmk server as part of the agent output. For example (cannot say if the syntax is 100% correct right now):
`
Write-Output (ConvertTo-Json -InputObject @{ command_creation_error = Out-String $Error[0]; })
`
To make progress here, we should split this up into more granular steps. We can also work on this together if things are unclear. Please ask if you need more input. I propose the following steps to make progress:

- Remove the `while` loop (see your ticket) and the current stdout and stderr logging
- Wrap the command creation into a `try`-`catch`-block (we can simply catch all errors here). In case of an error, we 1) log the error, 2) write the error to `$outputDirectory/scheduler_runner.json`, 3) exit.
- Introduce the environment build step. Without RCC, we can skip this step. With RCC, we build the command (including the same error handling as above) and execute. When executing, we collect stdout and stderr. In case of a failure, we report the error (logging + json) and exit.

Btw, to push to GitHub while rebasing, you can do something like this:
`git push -f --set-upstream origin HEAD:refs/heads/<branch name>`
This way, you can create smaller PRs but still use a single branch with multiple commits locally.

PR #326, opened 2023-09-05
RCC building: Use actions/checkout@v4

jherbel:
This might avoid the error described in
https://github.com/orgs/community/discussions/26321

However, we haven't encountered these errors in the RCC build up to now.


SoloJacobs (approved):
LGTM

PR #327, opened 2023-09-06
Collector: Report configuration or configuration reading error

jherbel:
CMK-14435

### DIFF
--- a/checkmk_extensions/agent/robotmk_collector.ps1
+++ b/checkmk_extensions/agent/robotmk_collector.ps1
@@ -6,23 +6,52 @@ function Main {
         [parameter(Mandatory=$true, Position=0)]
         [string]$ConfigPath
     )
-
-    $resultsDirectory = GetResultsDirectory($ConfigPath)
-    $files = Get-ChildItem -File -Recurse $resultsDirectory
-
     Write-Output "<<<robotmk_v2:sep(10)>>>"
+
+    try {
+        $configFileContent = Get-Content -Raw -Path $ConfigPath
+    }
+    catch {
+        Write-Output (SerializeConfigReadingError($Error[0]))
+        Throw $Error[0]
+    }
+
+    # We don't know if the config is actually valid, which is why don't simply dump it as is but
+    # instead wrap it.
+    Write-Output (SerializeConfigFileContent($configFileContent))
+
+    $files = Get-ChildItem -File -Recurse (GetResultsDirectory($configFileContent))
     foreach ($file in $files) {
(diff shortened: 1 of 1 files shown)


checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -6,23 +6,52 @@ function Main {
...
+    Write-Output "<<<robotmk_v2:sep(10)>>>"
 
-    $resultsDirectory = GetResultsDirectory($ConfigPath)
-    $files = Get-ChildItem -File -Recurse $resultsDirectory
+    try {
+        $configFileContent = Get-Content -Raw -Path $ConfigPath
+    }
+    catch {
+        Write-Output (SerializeConfigReadingError($Error[0]))
+        Throw $Error[0]
```
SoloJacobs, line 16: OTI: I guess this changes the exit code, but the windows agent still produces the output? Do we really want exit code 1.

checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -1,21 +1,30 @@
...
 }
 
-$resultsDirectory = GetResultsDirectory("${env:MK_CONFDIR}\robotmk.json")
-$files = Get-ChildItem -File -Recurse $resultsDirectory
-
-"<<<robotmk_v2:sep(10)>>>"
-foreach ($file in $files) {
-	Get-Content -ErrorAction 'SilentlyContinue' $file.FullName
-}
+Main("${env:MK_CONFDIR}\robotmk.json")
```
SoloJacobs, line 59: OTI: This is an unconventional signature for a main function.

SoloJacobs (approved):
LGTM

checkmk_extensions/agent/robotmk_collector.ps1
jherbel: Yes, the Windows agent still produces output if the plugin exits with 1, just tried it out. We already had exit code 1 before this change. `GetResultsDirectory` can throw an exception (for example if `results_directory` is missing or if the configuration isn't valid JSON).

checkmk_extensions/agent/robotmk_collector.ps1
jherbel: Yes, that's true. It would definitely be more standard to pass command line arguments (such as a path to a config file) to `Main`, but agent plugins are always called without any arguments. Would you prefer having `"${env:MK_CONFDIR}\robotmk.json"` inside `Main`?

SoloJacobs: I think it might be easier to test this way, so no =)

PR #330, opened 2023-09-12
Move config to separate module

jherbel:


SoloJacobs (approved):
LGTM

Just a thought for later: Since we want to parse the config inside of Checkmk, it make sense to have the schema as part of the API.

PR #332, opened 2023-09-13
Brush up UserSession

jherbel:


SoloJacobs (commented):
Do we really need `_TaskScheduler` and a `_Task` class? Perhaps a module would be more sensible here.

I also feel that `exit_code_path` is an implementation detail of our task scheduling and thus belongs into that interface therefore.

v2/robotmk/session.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
SoloJacobs, line 1: This feature will perhaps never make it into python, better just reorder the class definitions and remove this line.

SoloJacobs (changes requested):

PR #341, opened 2023-09-18
Attempt module: make traits used in tests only conditional

jherbel:


SoloJacobs (approved):
Is there any need to be this specific about these traits? It seems to me that as a consumer of the module, it would be very unfortunate if they are missing? And no downside to generating them?

jherbel:
I am currently not sure how to handle this, especially since there does not seem to be an easy for checking for unused implementations (which don't hurt though). For now, I would suggest to be strict about it (in the agent controller, we didn't do this). Maybe we can get a better idea how we should handle this later on. Reducing the strictness later on is very easy.

PR #342, opened 2023-09-18
Rust scheduler: Set up logging according to CLI options

jherbel:
CMK-14517

### DIFF
--- a/v2/rust/Cargo.toml
+++ b/v2/rust/Cargo.toml
@@ -8,3 +8,5 @@ edition = "2021"
 [dependencies]
 anyhow = { version = "*", features = ["backtrace"] }
 clap = { version = "*", features = ["derive"] }
+flexi_logger = "*"
+log = "*"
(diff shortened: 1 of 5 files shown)


v2/rust/src/cli.rs
```diff
@@ -17,3 +18,13 @@ pub struct Args {
     #[arg(short, long, action = ArgAction::Count)]
     pub verbose: u8,
```
SoloJacobs, line 19: should not be public?

SoloJacobs (dismissed):

v2/rust/src/cli.rs
jherbel: Done

v2/rust/Cargo.toml
```diff
@@ -8,3 +8,5 @@ edition = "2021"
 [dependencies]
 anyhow = { version = "*", features = ["backtrace"] }
 clap = { version = "*", features = ["derive"] }
+flexi_logger = "*"
```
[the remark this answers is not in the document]
jherbel: I think it should do also parallel execution, but we'll have to see that. Otherwise, it's simple to use and offers exactly the options we need (so far).

PR #343, opened 2023-09-18
Dev/sol/parse

SoloJacobs:


v2/rust/Cargo.toml
```diff
@@ -7,4 +7,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = { version = "*", features = ["backtrace"] }
+chrono = "0.4.31"
```
jherbel, line 10: Do we need to pin this atm? And also the other dependencies below.

jherbel (commented):
Only a question regarding the depedencies

v2/rust/Cargo.toml
SoloJacobs: By pin you probably mean pinning the major version and requiring a lower version? I was just following the recommendations here: https://doc.rust-lang.org/cargo/reference/resolver.html#recommendations

jherbel (approved):

PR #344, opened 2023-09-20
Rust scheduler: Config specification and parsing

jherbel:
CMK-14520

### DIFF
--- a/v2/data/retry_rcc/windows.json
+++ b/v2/data/retry_rcc/windows.json
@@ -1,28 +1,27 @@
 {
-	"working_directory": "/tmp/outputdir",
-	"results_directory": "/tmp/results",
-	"log_directory": "/tmp/logs",
-	"environment": {
-		"rcc_binary_path": "C:\\Users\\vagrant\\AppData\\Local\\Microsoft\\WindowsApps\\rcc.exe",
-		"scheduler_robot_yaml_path": "C:\\robotmk\\v2\\data\\scheduler_rcc\\robot.yaml"
-	},
+	"working_directory": "C:\\tmp\\output",
+	"results_directory": "C:\\tmp\\results",
 	"suites": {
 		"test": {
-			"execution_interval_seconds": 10,
-			"robot_target": "C:\\robotmk\\v2\\data\\retry_suite\\/tasks.robot",
-			"variants": [
-				{
-					"variablefile": null,
-					"argumentfile": null
-				},
-				{
-					"variablefile": "C:\\robotmk\\v2\\data\\retry_suite\\/retry_variables.yaml",
-					"argumentfile": null
-				}
-			],
-			"retry_strategy": "incremental",
-			"robot_yaml_path": "C:\\robotmk\\v2\\data\\retry_rcc\\/robot.yaml",
-			"session": null
+			"robot_framework_config": {
+				"robot_target": "C:\\robotmk\\v2\\data\\retry_suite\\tasks.robot",
+				"variable_file": "C:\\robotmk\\v2\\data\\retry_suite\\retry_variables.yaml",
+				"argument_file": null,
+				"retry_strategy": "Incremental"
+			},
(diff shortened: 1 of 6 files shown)


v2/rust/src/config.rs
```diff
@@ -0,0 +1,80 @@
...
+
+#[derive(Deserialize)]
+pub struct ExecutionConfig {
+    pub n_retries_max: usize,
+    pub execution_interval_seconds: u64,
+    pub timeout: u64,
+}
+
+#[derive(Deserialize)]
+#[serde(tag = "type")]
```
jherbel, line 48: This simplifies serialization in Python

v2/data/retry_rcc/windows.json
```diff
@@ -1,28 +1,27 @@
...
-					"variablefile": "C:\\robotmk\\v2\\data\\retry_suite\\/retry_variables.yaml",
-					"argumentfile": null
-				}
-			],
-			"retry_strategy": "incremental",
-			"robot_yaml_path": "C:\\robotmk\\v2\\data\\retry_rcc\\/robot.yaml",
-			"session": null
+			"robot_framework_config": {
+				"robot_target": "C:\\robotmk\\v2\\data\\retry_suite\\tasks.robot",
+				"variable_file": "C:\\robotmk\\v2\\data\\retry_suite\\retry_variables.yaml",
```
SoloJacobs, line 8: How this translate into the use case, which we discussed with Simon:
After trying a test, it should be retried with a different set of variable files?

v2/rust/src/config.rs
```diff
@@ -0,0 +1,80 @@
...
+    SpecificUser(UserSessionConfig),
+}
+
+#[derive(Deserialize)]
+pub struct UserSessionConfig {
+    pub user_name: String,
+}
+
+impl Config {
+    pub fn suites(&self) -> Vec<(&String, &SuiteConfig)> {
```
SoloJacobs, line 75: It's very difficult to review this, since it is not used anywhere.

SoloJacobs (changes requested):

v2/data/retry_rcc/windows.json
jherbel: Apparently, we misunderstood. This is not a use case we want to support.

v2/rust/src/config.rs
jherbel: It will be used in follow-up commits. The purpose is to make sure that the scheduler always builds environments in the same order. Would you prefer if I added this later on?

SoloJacobs: OTI: Interesting, but I guess the building of the environments is part of the test run? I guess this ensures the tests are started in the same order? The building the hololib library can be ensured to be somewhat deterministic in this manner I suppose.
If this is business logic, then it maybe warrants a test.

v2/data/retry_rcc/windows.json
SoloJacobs: I am actually very much in favor of this. Retries should not really be done with a different configuration :+1:

SoloJacobs (dismissed):

v2/rust/src/config.rs
jherbel: I added a test

v2/rust/src/config.rs
```diff
@@ -0,0 +1,133 @@
...
+            results_directory: PathBuf::from("/results"),
+            suites: HashMap::from([
+                (String::from("suite_b"), create_suite_config("b")),
+                (String::from("suite_a"), create_suite_config("a")),
+            ]),
+        }
+    }
+
+    #[test]
+    fn test_suites_sorted() {
```
SoloJacobs, line 124: Test name/documentation needs improving. Otherwise, the purpose of the test is unclear.

SoloJacobs (dismissed):
LGTM w/nit

PR #347, opened 2023-09-22
Rust scheduler: supervised child process execution

jherbel:
CMK-1456

### DIFF
--- a/v2/rust/src/child_process_supervisor.rs
+++ b/v2/rust/src/child_process_supervisor.rs
@@ -1,0 +1,103 @@
+use super::termination::TerminationFlag;
+use anyhow::{bail, Context, Result};
+use async_std::{future::timeout, task::sleep};
+use futures::executor;
+use log::{debug, error};
+use std::collections::{HashMap, HashSet};
+use std::process::{Child, Command, ExitStatus};
+use std::time::Duration;
+use sysinfo::{Pid, PidExt, Process, ProcessExt, System, SystemExt};
+
+pub struct ChildProcessSupervisor<'a> {
+    pub command: Command,
+    pub timeout: u64,
+    pub termination_flag: &'a TerminationFlag,
+}
+
+impl ChildProcessSupervisor<'_> {
+    pub fn run(mut self) -> Result<ChildProcessOutcome> {
+        debug!("Executing {:?}", self.command);
+        let mut child = self.command.spawn().context("Failed to spawn subprocess")?;
+        match executor::block_on(timeout(
+            Duration::from_secs(self.timeout),
+            self.wait_for_child_exit(&mut child),
+        )) {
+            Ok(child_result) => child_result,
+            _ => {
+                error!("Timed out");
+                kill_process_tree(&mut child);
+                Ok(ChildProcessOutcome::TimedOut)
+            }
+        }
+    }
+
+    async fn wait_for_child_exit(self, child: &mut Child) -> Result<ChildProcessOutcome> {
+        loop {
+            if let Some(exit_status) = child
+                .try_wait()
+                .context(format!(
+                    "Failed to query exit status of process {}, killing",
+                    child.id()
+                ))
+                .map_err(|err| {
+                    kill_process_tree(child);
+                    err
+                })?
+            {
+                return Ok(ChildProcessOutcome::Exited(exit_status));
+            }
+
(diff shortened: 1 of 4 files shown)


v2/rust/src/child_process_supervisor.rs
```diff
@@ -0,0 +1,93 @@
...
+use std::time::Duration;
+use sysinfo::{Pid, PidExt, Process, ProcessExt, System, SystemExt};
+
+pub struct ChildProcessSupervisor<'a> {
+    pub command: Command,
+    pub timeout: u64,
+    pub termination_flag: &'a TerminationFlag,
+}
+
+impl ChildProcessSupervisor<'_> {
```
SoloJacobs, line 17: A timeout functionality? It feels like this should already somewhere? This would be an example:
https://docs.rs/async-std/latest/async_std/future/fn.timeout.html

v2/rust/src/child_process_supervisor.rs
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            if timeout.expired() {
+                error!("Process timed out");
+                kill_child_process_tree(&mut child);
+                return Ok(ChildProcessOutcome::TimedOut);
+            }
+
+            if self.termination_flag.should_terminate() {
+                warn!("Terminated");
```
SoloJacobs, line 44: Why have the same message twice?

v2/rust/src/child_process_supervisor.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+}
+
+pub enum ChildProcessOutcome {
+    Exited(ExitStatus),
+    TimedOut,
+}
+
+fn kill_child_process_tree(child: &mut Child) {
```
SoloJacobs, line 58: `kill_child_and_process_tree` and `kill_all_children` feel like the same function. Consider combining them into one function.

v2/rust/src/child_process_supervisor.rs
```diff
@@ -0,0 +1,93 @@
...
+    let mut system = System::new_all();
+    system.refresh_processes();
+    let _ = child.kill();
+    kill_all_children(&Pid::from_u32(child.id()), system.processes());
+}
+
+fn kill_all_children<'a>(top_pid: &'a Pid, processes: &'a HashMap<Pid, Process>) {
```
SoloJacobs, line 65: Is this really the correct behaviour? From what I can see there are three things somebody can do:
* Send a SIGKILL signal such as this:
```
taskkill /pid 1230 /f
```
In such a case, `kill_all_children` would never be invoked, correct?
* Send a SIGKILL signal such as this:
```
taskkill /pid 1230 /f /t 
```
Again, taskkill is not invoked, but this time it doesn't matter, since all children are terminated by the invoker of `taskkill`.
* Send a SIGTERM, such as doing control-C on the command line. In this case, we need to invoking SIGTERM on each direct child only should be the correct behaviour. We then need to understand how rcc and robotframework handle such terminations.

SoloJacobs (commented):
Some questions

v2/rust/src/child_process_supervisor.rs
jherbel: Done

v2/rust/src/child_process_supervisor.rs
jherbel: Or maybe we could rename `kill_child_process_tree(child: &mut Child)` to `kill_process_tree(root_process: &mut Child)`?

v2/rust/src/child_process_supervisor.rs
jherbel: Windows simply doesn't have `SIGTERM`, I also confirmed this with Sergey. Otherwise, `SIGTERM` would be the way to go. `taskkill /t` does not work for our use cases, I think it only works with foreground applications, not sure though. It definitely cannot kill a Python process, I tried that.

v2/rust/src/child_process_supervisor.rs
jherbel: Done

v2/rust/src/child_process_supervisor.rs
jherbel: As discussed via Slack: I added documentation.

PR #350, opened 2023-09-26
Rust scheduler: environment building

jherbel:
CMK-14557

### DIFF
--- a/v2/rust/src/environment.rs
+++ b/v2/rust/src/environment.rs
@@ -1,5 +1,234 @@
+use super::child_process_supervisor::{ChildProcessOutcome, ChildProcessSupervisor};
+use super::config::{Config, EnvironmentConfig};
+use super::results::{EnvironmentBuildStatesAdministrator, EnvironmentBuildStatus};
+use super::termination::TerminationFlag;
+use anyhow::{Context, Result};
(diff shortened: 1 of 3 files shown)


v2/rust/src/environment.rs
```diff
@@ -1,5 +1,234 @@
...
+            Some(mut build_instructions) => {
+                info!("Building environment for suite {}", suite_name);
+                environment_build_states_administrator
+                    .insert_and_write_atomic(suite_name, EnvironmentBuildStatus::InProgress)?;
+                configure_stdio_of_environment_build(
+                    &env_building_stdio_directory,
+                    suite_name,
+                    &mut build_instructions.command,
+                )
+                .context("Configuring stdio of environment build process failed")?;
```
SoloJacobs, line 39: OTI: Is this really correct? I think failing to build one environment should not affect the remaining tests. Better build everything, otherwise the user has to reconfigure till next error.

jherbel: I am not 100% sure about it either. However, if this happens, there is anyway a severe problem (we were unable to open files). So most likely, it doesn't matter if we crash here or later on, since nothing would run anyway. If this turns out to be a problem, we can change it.

PR #367, opened 2023-10-09
Adjust working directory and RCC binary permissions during setup

jherbel:
For headful tests, we need to give the corresponding user the permissions to
* read and write to the working directory
* execute the RCC binary (if configured)

### DIFF
--- a/v2/rust/src/setup.rs
+++ b/v2/rust/src/setup.rs
@@ -1,12 +1,14 @@
-use anyhow::{Context, Result};
-use atomicwrites::replace_atomic;
-use camino::{Utf8Path, Utf8PathBuf};
-use std::collections::HashSet;
-use std::fs::{create_dir_all, remove_file};
-
-use super::config::Config;
+use super::config::{Config, EnvironmentConfig};
 use super::environment::environment_building_stdio_directory;
 use super::results::{suite_result_file, suite_results_directory};
+
+use anyhow::{bail, Context, Result};
+use atomicwrites::replace_atomic;
+use camino::{Utf8Path, Utf8PathBuf};
+use log::debug;
+use std::collections::HashSet;
+use std::fs::{create_dir_all, remove_file};
+use std::process::Command;
 
 pub fn setup(config: &Config) -> Result<()> {
     create_dir_all(&config.working_directory).context("Failed to create working directory")?;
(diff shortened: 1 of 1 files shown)


v2/rust/src/setup.rs
```diff
@@ -17,7 +20,9 @@ pub fn setup(config: &Config) -> Result<()> {
     create_dir_all(&config.results_directory).context("Failed to create results directory")?;
     create_dir_all(suite_results_directory(&config.results_directory))
         .context("Failed to create suite results directory")?;
-    clean_up_results_directory_atomic(config)
+    clean_up_results_directory_atomic(config)?;
+    adjust_user_permissions(config);
```
SoloJacobs, line 24: Should we not pass these errors?

jherbel: We have two options here:
1. Pass the error. Since we have no good way of handling the error, our setup would simply stop and pass the error to the main function.
2. Log the error and continue with the next suite.

I vote for option 2 (of course Simon should ultimately decide this). It's true that all other steps above immediately lead to an abort. However, errors in these steps indicate I/O issues, which are quite severe, so I think an abort is justified here. On the other hand, I am not sure how severe an error in the permissions step is.

v2/rust/src/setup.rs
```diff
@@ -72,3 +77,63 @@ fn remove_files_atomic<'a>(
...
+    let user_name = match &suite_config.session_config {
+        SessionConfig::SpecificUser(user_session_config) => &user_session_config.user_name,
+        _ => return Ok(()),
+    };
+
+    debug!("Granting {user_name} full access to {working_directory}");
+    run_icacls_command(vec![
+        working_directory.as_str(),
+        "/grant",
+        &format!("{user_name}:(OI)(CI)F"),
```
SoloJacobs, line 102: I don't understand what this list means?

SoloJacobs: Are containers just directories and objects just files?

SoloJacobs (dismissed):
Not pretty, but it's windows

v2/rust/src/setup.rs
SoloJacobs: Hmm, seems like we should maybe at least forward this error to Checkmk?

jherbel: After talking to Simon, we decided to do this a bit differently: Instead of setting permissions for individual users, we set permissions for the group `Users`. And if this fails, the scheduler is allowed to crash.

v2/rust/src/setup.rs
jherbel: I think so, but I couldn't find any official documentation on this. Windows ...

PR #374, opened 2023-10-11
Add possibility to provide path as argument and default value 

racicLuka:
Added possibility to call the script with an argument:
`./robotmk_collector.ps1 "path_as_argument"`

Also, if `env:MK_CONFDIR` is not present, the path will default to `C:\ProgramData\checkmk\agent\config\robotmk.json`

### DIFF
--- a/checkmk_extensions/agent/robotmk_collector.ps1
+++ b/checkmk_extensions/agent/robotmk_collector.ps1
@@ -2,10 +2,7 @@ Set-StrictMode -Version 3.0
 $ErrorActionPreference = "Stop"
 
 function Main {
-    Param (
-        [parameter(Mandatory=$true, Position=0)]
-        [string]$ConfigPath
-    )
+    $ConfigPath = DetermineConfigPath -CommandLineArgs $args
     Write-Output "<<<robotmk_v2:sep(10)>>>"
 
     try {
@@ -27,6 +24,24 @@ function Main {
 }
 
 
+function DetermineConfigPath {
+    [OutputType([string])]
+    Param (
+        [parameter(Mandatory=$false, Position=0)]
+        [String[]]$CommandLineArgs
+    )
+
+    if ($CommandLineArgs -ne "" -and $CommandLineArgs.Count -gt 0) {
+        return $CommandLineArgs[0]
+    }
+
+    $configDir = $env:MK_CONFDIR
+    if ($null -eq $configDir) {
+        $configDir = 'C:\ProgramData\checkmk\agent\config'
+    }
+
+    return Join-Path $configDir 'robotmk.json'
+}
 function SerializeConfigReadingError {
     [OutputType([string])]
     Param (
@@ -56,4 +71,4 @@ function GetResultsDirectory {
     return $configData.results_directory -as [string]
 }
 
-Main("${env:MK_CONFDIR}\robotmk.json")
+Main $args


checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -3,9 +3,26 @@ $ErrorActionPreference = "Stop"
 
 function Main {
     Param (
-        [parameter(Mandatory=$true, Position=0)]
+        [parameter(Mandatory=$false, Position=0)]
         [string]$ConfigPath
     )
+
+    #If the path is provided as an argument, use it.
```
jherbel, line 10: Let's move the new code into a separate function

checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -3,9 +3,26 @@ $ErrorActionPreference = "Stop"
 
 function Main {
     Param (
-        [parameter(Mandatory=$true, Position=0)]
+        [parameter(Mandatory=$false, Position=0)]
```
jherbel, line 6: Since we are never calling `Main` with a parameter, we can remove it alltogether

checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -3,9 +3,26 @@ $ErrorActionPreference = "Stop"
...
+
+    #If the path is provided as an argument, use it.
+    if ($args.Count -gt 0) {
+        $ConfigPath = $args[0]
+    }
+
+    # If the -ConfigPath parameter is not defined, check the env:MK_CONFDIR environment variable.
+    # If it is not defined, default to C:\ProgramData\checkmk\agent\config\robotmk.json.
+    if (!$ConfigPath) {
+        $ConfigPath = $env:MK_CONFDIR
```
jherbel, line 18:
```
$configDir = $env:MK_CONFDIR
if (!$configDir) {
    $configDir = 'C:\ProgramData\checkmk\agent\config'
}
return $configDir + '\robotmk.json'
```

jherbel (changes requested):
Let's also remove the comments, they shouldn't be necessary

checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -26,7 +23,26 @@ function Main {
     }
 }
 
+function DetermineConfigPath {
+    [OutputType([string])]
+    param (
+        [array]$ProvidedPath
```
jherbel, line 29: I think this should be `String []`

checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -26,7 +23,26 @@ function Main {
     }
 }
 
+function DetermineConfigPath {
+    [OutputType([string])]
+    param (
```
jherbel, line 28:
```
Param (
        [parameter(Mandatory=$true, Position=0)]
```

checkmk_extensions/agent/robotmk_collector.ps1
jherbel, line 29: Let's call it `$CommandLineArgs` or so, since we don't know yet if there actually is a path

checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -56,4 +72,4 @@ function GetResultsDirectory {
     return $configData.results_directory -as [string]
 }
 
-Main("${env:MK_CONFDIR}\robotmk.json")
+Main $args
```
jherbel, line 75: This is not needed

jherbel (changes requested):
Please remove the code comments and squash the PR into a single commit

racicLuka:
I can squash and merge once you approve the PR, if that is OK with you?

jherbel:
> I can squash and merge once you approve the PR, if that is OK with you?

Depends on how the commit would look afterwards. The easiest way would be to simply squash locally and do a force push.

checkmk_extensions/agent/robotmk_collector.ps1
jherbel: What I meant:
`[String []]$CommandLineArgs`
So an array of strings

jherbel (commented):
See the last remaining comment about `[String []]$CommandLineArgs`

checkmk_extensions/agent/robotmk_collector.ps1
```diff
@@ -27,6 +24,24 @@ function Main {
...
 
 
+function DetermineConfigPath {
+    [OutputType([string])]
+    Param (
+        [parameter(Mandatory=$false, Position=0)]
+        [String[]]$CommandLineArgs
+    )
+
+    if ($CommandLineArgs -ne "" -and $CommandLineArgs.Count -gt 0) {
```
racicLuka, line 34: I needed to change this to `$CommandLineArgs -ne ""` because now, if there are no arguments provided, it considers it as an empty string.

jherbel (approved):

PR #376, opened 2023-10-12
Archive Python prototype scheduler

jherbel:
Also remove it from the CI.


archive/v2/tests/rebot.xml
SoloJacobs, line 1: This seems to be wrong. The file here is actually used by the rust tests. This commit does not move, it only copies.

jherbel: I copied it on purpose, st. the tests of the Python scheduler could in principle still run. In a follow-up commit, I will move the original file into the tests directory of the scheduler

SoloJacobs (commented):
Why should also consider whether we simply delete this... But we can postpone this, after we have decideed what to do with v1

jherbel:
> Why should also consider whether we simply delete this... But we can postpone this, after we have decideed what to do with v1

Agreed, but for now, I let's archive it. We might still need this as a reference for some time.

SoloJacobs (approved):
LGTM

PR #379, opened 2023-10-12
Rename Rust scheduler directory and move tests directory below it

jherbel:


SoloJacobs (commented):
It this robotmk part really necessary? I think we only did that because of python weirdness.

jherbel:
It's not necessary, but I find it strange to keep the directory name "rust", even though there is nothing else in there. And at some point, we will anyway remove the v2-layer.

SoloJacobs (dismissed):

SoloJacobs (approved):

PR #380, opened 2023-10-12
chore: sign cla as RobWalt

RobWalt:
Add myself to list of signers

Let me know if I need to do it in a different way


SoloJacobs:
The signature is the GitHub comment, rather than this file (which is merely a cache for bookkeeping). The problem with using this branch as "the signature" is that we could arbitrarily rewrite the history. Thus, we really need the GitHub comment. I don't know why it didn't trigger correctly. I will have to debug this and let you know.

PR #383, opened 2023-10-16
Decouple `Session` and `Environment`

jherbel:
This makes it easier to split the configuration into a global configuration and per-suite configuration.

Furthermore, originally, the idea was that `Session` takes a command as input, wraps it using `Environment` and then runs the wrapped command. However, the requirement that the scheduler configures shared holotrees means that `Session` must also be able to run un-wrappend commands (`rcc holotree init`). So decoupling these two also paves the way for implementing this requirement.

CMK-14802

### DIFF
--- a/v2/robotmk/src/session.rs
+++ b/v2/robotmk/src/session.rs
@@ -1,67 +1,60 @@
-use super::attempt::Attempt;
 use super::child_process_supervisor::{ChildProcessOutcome, ChildProcessSupervisor, StdioPaths};
+use super::command_spec::CommandSpec;
 use super::config::SessionConfig;
-use super::environment::{Environment, ResultCode};
 use super::termination::TerminationFlag;
-use anyhow::Result;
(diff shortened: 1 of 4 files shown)


v2/robotmk/src/session.rs
```diff
@@ -1,67 +1,60 @@
...
         }
     }
 }
 
-pub struct CurrentSession<'a> {
-    environment: &'a Environment,
-    termination_flag: &'a TerminationFlag,
+pub struct CurrentSession {}
+
+pub struct RunSpec<'a> {
```
jherbel, line 30: I'm open to naming suggestions...

SoloJacobs: My understand is, that this field is unused?

jherbel: Yes, `id` is currently unused, see my comment above.

PR #384, opened 2023-10-17
Implement and use internal configuration format

jherbel:
CMK-14802


v2/robotmk/src/config/external.rs
```diff
@@ -49,88 +48,28 @@ pub struct ExecutionConfig {
...
-#[derive(Clone, Deserialize)]
-#[cfg_attr(test, derive(Debug, PartialEq))]
+#[derive(Deserialize)]
 pub struct RCCEnvironmentConfig {
     pub binary_path: Utf8PathBuf,
     pub robot_yaml_path: Utf8PathBuf,
     pub build_timeout: u64,
 }
 
-#[derive(Clone, Deserialize)]
```
SoloJacobs: FYI: Robert mentioned that he considers leaving some nice to have traits as good practice, albeit this is of course completely up the team (except maybe if binary size is a concern).

v2/robotmk/src/attempt.rs
```diff
@@ -215,7 +213,7 @@ mod tests {
                 name: "suite_1",
                 timestamp: "2023-08-29T12.23.44.419347+00.00".into(),
             },
-            working_directory: &Utf8PathBuf::from("/tmp/outputdir/"),
+            working_directory: &Utf8PathBuf::from("/tmp/outputdir/suite_1"),
```
SoloJacobs, line 216: What does this change mean?

v2/robotmk/src/config/external.rs
jherbel: Ok, thanks for asking. Maybe we should review this at some point and decide on a standard set of traits which we always derive if possible.

v2/robotmk/src/attempt.rs
jherbel: `RetrySpec` does not add the suite name anymore when computing `output_directory`, since this already part of `Suite.working_directory` and handed over to `RetrySpec.working_directory`. That's why I added it here.

PR #389, opened 2023-10-20
Implement headed tests

jherbel:


v2/robotmk/src/session.rs
```diff
@@ -10,19 +11,25 @@ use camino::{Utf8Path, Utf8PathBuf};
 #[cfg_attr(test, derive(Debug, PartialEq))]
 pub enum Session {
     Current(CurrentSession),
+    SpecificUser(UserSession),
```
SoloJacobs, line 14: I would shorten this to `User`

v2/robotmk/src/session.rs
```diff
@@ -67,3 +80,16 @@ impl CurrentSession {
         }
     }
 }
+
+impl UserSession {
+    fn run(&self, spec: &RunSpec) -> Result<RunOutcome> {
+        run_task(&TaskSpec {
```
SoloJacobs, line 86: Should this not be inlined?

v2/robotmk/src/schtasks.rs
```diff
@@ -0,0 +1,265 @@
...
+        }
+    };
+    let pid = match Pid::from_str(&raw_pid).context(format!("Failed to parse {} as PID", raw_pid)) {
+        Ok(pid) => pid,
+        Err(err) => {
+            log_and_return_error(err);
+            return;
+        }
+    };
+    kill_process_tree(&pid);
```
SoloJacobs, line 207: I assume you have tested, that windows does not terminate the task processes?

v2/robotmk/src/schtasks.rs
```diff
@@ -0,0 +1,265 @@
...
+        Err(err) => {
+            log_and_return_error(err);
+            return;
+        }
+    };
+    let pid = match Pid::from_str(&raw_pid).context(format!("Failed to parse {} as PID", raw_pid)) {
+        Ok(pid) => pid,
+        Err(err) => {
+            log_and_return_error(err);
+            return;
```
SoloJacobs, line 204: This would be worth a panic.

v2/robotmk/src/schtasks.rs
```diff
@@ -0,0 +1,265 @@
...
+    error!("Killing and deleting task {task_name}");
+    kill_task(path_pid);
+    delete_task(task_name);
+}
+
+fn kill_task(path_pid: &Utf8Path) {
+    let raw_pid = match read_from_file_before_first_whitespace(path_pid) {
+        Ok(raw_pid) => raw_pid,
+        Err(err) => {
+            log_and_return_error(err);
```
SoloJacobs, line 196: This would be worth a panic

v2/robotmk/src/schtasks.rs
```diff
@@ -0,0 +1,265 @@
...
+
+fn query_if_task_is_running(task_name: &str) -> Result<bool> {
+    let schtasks_stdout = run_schtasks(["/query", "/tn", task_name, "/fo", "CSV", "/nh"])?;
+    Ok(schtasks_stdout.contains("Running"))
+}
+
+fn kill_and_delete_task(task_name: &str, path_pid: &Utf8Path) {
+    error!("Killing and deleting task {task_name}");
+    kill_task(path_pid);
+    delete_task(task_name);
```
SoloJacobs, line 189: Does order matter here?

v2/robotmk/src/schtasks.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+fn delete_task(task_name: &str) {
+    debug!("Deleting task {task_name}");
+    let _ = run_schtasks(["/delete", "/tn", task_name, "/f"])
+        .context(format!("Failed to delete task {}", task_name))
+        .map_err(log_and_return_error);
+}
+
+fn read_from_file_before_first_whitespace(path: &Utf8Path) -> Result<String> {
```
SoloJacobs, line 217: Confusing, test? Maybe call this read_until_first_whitespace. The path makes the file part clear enough

v2/robotmk/src/schtasks.rs
```diff
@@ -0,0 +1,265 @@
...
+use chrono::{Duration as ChronoDuration, Local};
+use futures::executor;
+use log::{debug, error};
+use std::fs::{read_to_string, write};
+use std::process::Command;
+use std::str::FromStr;
+use std::time::Duration;
+use sysinfo::Pid;
+
+pub fn run_task(task_spec: &TaskSpec) -> Result<RunOutcome> {
```
SoloJacobs, line 18: Could we try having explicit errors here?

SoloJacobs (commented):
LGTM (aside from windows ugliness), just a few remarks. No must-have changes, but worth discussing.

v2/robotmk/src/schtasks.rs
jherbel: As discussed, this is not the right place to try this out

v2/robotmk/src/schtasks.rs
jherbel: Done

v2/robotmk/src/schtasks.rs
jherbel: No, I don't think so

v2/robotmk/src/schtasks.rs
jherbel: Yes. schtasks /end terminates the task itself, but any child processes survive ... I left a comment explaining this.

v2/robotmk/src/session.rs
jherbel: I decided to leave it as is. We would otherwise have to expose some implementation details from schtasks.rs, so I would rather leave it as is.

v2/robotmk/src/session.rs
jherbel: Done

v2/robotmk/src/schtasks.rs
jherbel: Yes, we could indeed panic here. Let's leave it as is for now, we can always change it if necessary. At least, the error isn't completely lost, since we log it.

v2/robotmk/src/schtasks.rs
jherbel: As we discussed, there is a race condition here. I made this a bit more robust by adding a one-time sleep in case of a failure.

SoloJacobs (approved):

PR #391, opened 2023-10-23
Dev/sol/refactor ii

SoloJacobs:


v2/robotmk/src/environment.rs
```diff
@@ -65,16 +64,20 @@ fn build_environment<'a>(
             let drop_suite = matches!(environment_build_status, EnvironmentBuildStatus::Failure(_));
             environment_build_states_administrator
                 .insert_and_write_atomic(&suite.name, environment_build_status)?;
-            drop_suite
+            if drop_suite {
```
jherbel, line 67: I think we have to invert this, right? We want to return `Some(suite)` if we the building suceeded.

v2/robotmk/src/environment.rs
```diff
@@ -65,16 +64,20 @@ fn build_environment<'a>(
...
+            } else {
+                None
+            }
         }
         None => {
             debug!("Nothing to do for suite {}", suite.name);
             environment_build_states_administrator
                 .insert_and_write_atomic(&suite.name, EnvironmentBuildStatus::NotNeeded)?;
-            false
+            None
```
jherbel, line 77: See above

jherbel (approved):
Nice idea :+1: Just a small comment regarding the discard logic.

jherbel (changes requested):
Sorry, chose "Approve" instead of "Request changes" before

v2/robotmk/src/environment.rs
SoloJacobs: Yes, good catch. Not sure how to test this :thinking:

v2/robotmk/src/environment.rs
SoloJacobs: Done

jherbel (approved):

PR #392, opened 2023-10-23
Dev/sol/agent plugin

SoloJacobs:


v2/robotmk/src/bin/agent.rs
```diff
@@ -0,0 +1,139 @@
...
+    if let Ok(dir) = entry {
+        if dir.file_type().is_file() {
+            if let Ok(raw) = read_to_string(dir.path()) {
+                writeln!(stdout, "{raw}").unwrap();
+            }
+        }
+    }
+}
+
+fn walk(results_directory: &Path, stdout: &mut impl io::Write) {
```
jherbel, line 73: You could use `impl AsRef<Path>` here, then you could pass `&config.results_directory` at the call site

v2/robotmk/src/bin/agent.rs
```diff
@@ -0,0 +1,139 @@
...
+    for entry in WalkDir::new(results_directory)
+        .sort_by_file_name()
+        .into_iter()
```
jherbel, line 76: Since we are (correctly) ignoring errors in `print_or_ignore`, you could directly transform into an iterator which contains `DirEntry` by discarding any errors:
```
.filter_map(|dir_entry_result| dir_entry_result.ok())
```
Then you would have one nesting level less in `print_or_ignore`.

v2/robotmk/src/bin/agent.rs
```diff
@@ -0,0 +1,139 @@
...
+    let mut config_dir = Utf8PathBuf::from(config_dir);
+    config_dir.push("robotmk.json");
+    Ok(config_dir)
+}
+
+fn report_config_error(message: String) {
+    let config_error = serde_json::to_string(&ConfigError {
+        config_reading_error: message,
+    })
+    .unwrap();
```
jherbel, line 51: I would prefer to return a result here and unwrap in `main`, but I think that's more a question of taste, so no need to change it.

v2/robotmk/src/bin/agent.rs
```diff
@@ -0,0 +1,139 @@
...
+    use tempfile::tempdir;
+    // Assemble
+    let path_content = [
+        ("1", "Failure is not an Option<T>, it's a Result<T,E>."),
+        ("2", "In Rust, None is always an Option<_>."),
+        ("3/nested", "Rust is the best thing since &Bread[..]."),
+        ("4/more/nesting", "Yes, I stole these jokes from reddit."),
+    ];
+    let expected: String = path_content.map(|(_, c)| format!("{c}\n")).concat();
+    let results_directory = tempdir().unwrap();
```
jherbel, line 127: Is this directory cleaned up once `results_directory` goes out of scope? I think it should be, but I am not sure.

v2/robotmk/src/bin/agent.rs
```diff
@@ -0,0 +1,139 @@
...
+pub struct ConfigError {
+    config_reading_error: String,
+}
+
+#[derive(Serialize)]
+pub struct ConfigFileContent {
+    config_file_content: String,
+}
+
+fn determine_config_path(arg: Option<Utf8PathBuf>) -> Result<Utf8PathBuf, String> {
```
jherbel, line 33: Personally, I would have split this into two functions:
```
fn determine_config_path(arg: Option<Utf8PathBuf>) -> Result<Utf8PathBuf, String> {
    Ok(arg.unwrap_or(config_path_from_env()?)
}

fn config_path_from_env() -> Result<Utf8PathBuf, String> {
    Ok(Utf8PathBuf::from(
        match var("MK_CONFDIR") {...}
    ).join("robotmk.json"))
}
```
But I think this is more a question of taste and not a hard requirement from my side.

jherbel (dismissed):
Nice job, I have some minor optional suggestions. Check if you agree with them. If not, it's also ok for me to merge it as is.

jherbel (approved):

jherbel (approved):

PR #403, opened 2023-10-27
Add generic option for RF command line arguments to configuration

jherbel:
This replaces the existing options `argument_file` and `variable_file` and paves the way for allowing the user to configure further command line arguments in the bakery.

CMK-14941


racicLuka (approved):
LGTM

PR #410, opened 2023-10-30
Cleanup job for suite working directories

jherbel:


v2/robotmk/src/scheduling/cleanup.rs
```diff
@@ -0,0 +1,121 @@
...
+        .file_type()
+        .context(format!(
+            "Failed to determine file type of {}",
+            dir_entry.path()
+        ))?
+        .is_dir()
+    {
+        remove_dir_all(dir_entry.path()).context(format!("Failed to remove {}", dir_entry.path()))
+    } else {
+        remove_file(dir_entry.path()).context(format!("Failed to remove {}", dir_entry.path()))
```
SoloJacobs, line 119: Seems redundant?

v2/robotmk/src/scheduling/cleanup.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+fn dir_entries_to_remove_max_executions(
+    mut dir_entries: Vec<Utf8DirEntry>,
+    max_executions: usize,
+) -> Vec<Utf8DirEntry> {
+    if max_executions > dir_entries.len() {
+        return vec![];
+    }
+    dir_entries.sort_by_key(|dir_entry| dir_entry.path().to_path_buf());
```
SoloJacobs, line 102: They should be sorted by age, no?

v2/robotmk/src/scheduling/cleanup.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+fn dir_entries_to_remove_max_executions(
+    mut dir_entries: Vec<Utf8DirEntry>,
+    max_executions: usize,
+) -> Vec<Utf8DirEntry> {
+    if max_executions > dir_entries.len() {
+        return vec![];
+    }
```
SoloJacobs, line 101: redundant?

SoloJacobs (changes requested):

v2/robotmk/src/scheduling/cleanup.rs
jherbel: I rely on the timestamps in the working directory names for the sorting by age. I added a comment for this. Would you prefer an explicit sorting by age?

v2/robotmk/src/scheduling/cleanup.rs
jherbel: This was necessary to avoid panics at `dir_entries.len() - max_executions`. However, I re-formulated this part and added tests.

v2/robotmk/src/scheduling/cleanup.rs
SoloJacobs: This seems like unnecessary coupling to me. Moreover, I'm not convinced, that the two implementations behave the same in all edge cases.

v2/robotmk/src/scheduling/cleanup.rs
```diff
@@ -0,0 +1,194 @@
...
+        let _ = remove_dir_entry(&dir_entry).map_err(log_and_return_error);
+    }
+    Ok(())
+}
+
+fn dir_entries_to_remove_max_age(
+    dir_entries: Vec<Utf8DirEntry>,
+    max_age_secs: u64,
+) -> Vec<Utf8DirEntry> {
+    let now = SystemTime::now();
```
SoloJacobs, line 60: OTI: Should maybe log this.

SoloJacobs (approved):

PR #415, opened 2023-11-03
Rename `n_retries_max` --> `n_attempts_max`

jherbel:
The latter is more concise. `n_retries_max` could be interpreted as not including the first attempt (since the first one is not a retry but a "try").


SoloJacobs (approved):
LGTM

PR #419, opened 2023-11-06
Avoid race condition when cleaning up deprecated result files

jherbel:


SoloJacobs (changes requested):
I am not sure this is the correct approach. The agent plugin now reads and writes to the directory. The scheduler is still writing to the directory. This means there is still are race conditions: If plugin is reading from the directory, then the scheduler can fail to write his sections, since they are currently locked.

I understand that this approach can  alleviate the problem, because the execution of the file deletion now happens much more frequently. However, we could accomplish the same by having a frequently running deletion job in the scheduler. This way there are not multiple actors, who need write access to the results directory.

SoloJacobs (approved):

PR #420, opened 2023-11-06
Introduce `WriteSection` trait

SoloJacobs:


v2/robotmk/src/section.rs
```diff
@@ -0,0 +1,59 @@
...
+use tempfile::NamedTempFile;
+use walkdir::WalkDir;
+
+#[derive(Deserialize, Serialize)]
+pub struct Section {
+    pub name: String,
+    pub content: String,
+}
+
+fn write(name: String, content: &impl Serialize, path: impl AsRef<Utf8Path>) -> Result<()> {
```
jherbel, line 16: Is there a specific reason for having `name` and `content` as separate arguments? Why not simply pass a `Section`? That would also make the function symmetric wrt. to `read`.

jherbel (approved):
Looks good overall, I just had one minor question

v2/robotmk/src/section.rs
SoloJacobs: Yes, that is true at the moment. One could then also inline `write`. However, we need `write` for implementing piggyback hosts later: https://github.com/elabit/robotmk/pull/422/commits/f878caf0e31ebeb927d2828648dbeecc9f6cc38b
There then is choice between duplicating the Section creation logic and sharing it via the `write` function. We can explore it in a later commit, but I chose this implementation on purpose.

PR #425, opened 2023-11-07
Lock results directory before read or write access

jherbel:


SoloJacobs (approved):
LGTM, might be worth switching to sqlite in the future, but that's for another day.

PR #426, opened 2023-11-07
Minor fixes for sections

jherbel:
* Name all sections "robotmk_..."
* Use null character as separator
* Only report configuration if deserialization was succesful

### DIFF
--- a/v2/robotmk/src/bin/agent.rs
+++ b/v2/robotmk/src/bin/agent.rs
@@ -19,13 +19,9 @@ struct Args {
 }
 
 #[derive(Serialize)]
-pub struct ConfigError {
-    config_reading_error: String,
-}
-
-#[derive(Serialize)]
-pub struct ConfigFileContent {
-    config_file_content: String,
+pub enum ConfigSection {
+    ReadingError(String),
+    FileContent(String),
 }
 
 fn determine_config_path(arg: Option<Utf8PathBuf>) -> Result<Utf8PathBuf, String> {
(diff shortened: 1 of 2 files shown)


v2/robotmk/src/bin/agent.rs
```diff
@@ -51,13 +52,14 @@ fn report_config_content(content: String) {
         config_file_content: content,
     })
     .expect("Unexpected serialization error: ConfigFileContent");
+    println!("<<<robotmk_config:sep(0)>>>");
```
SoloJacobs, line 55: There should be different headers for different JSON schemas. Moreover, we should simplify the implementations here, so that the sections are always printed.

If we do not adjust these sections in tandem with commits on the Checkmk side, then it is  very hard to judge whether these changes simplify parsing.

v2/robotmk/src/bin/agent.rs
```diff
@@ -82,14 +83,14 @@ fn main() {
...
     };
-    report_config_content(raw.clone());
     let config: Config = match serde_json::from_str(&raw) {
         Ok(config) => config,
         Err(e) => {
             report_config_error(e.to_string());
             return;
         }
     };
+    report_config_content(raw.clone());
```
SoloJacobs, line 93: Having the config is particularly useful, if the de-serialization fails. Reporting the content if available was intentional on my end.

SoloJacobs (approved):
I don't mind changing the agent, but it's unclear to me that this is the correct approach.

v2/robotmk/src/bin/agent.rs
jherbel: As discussed: It's sufficient to report the config if the parsing suceeded.

v2/robotmk/src/bin/agent.rs
jherbel: As discussed: I unified the configuration into one section.

SoloJacobs (approved):
LGTM

PR #431, opened 2023-11-07
Dev/sol/artifacts ii

SoloJacobs:


jherbel (approved):
The title of the second commit has a typo

PR #433, opened 2023-11-09
Dev/sol/tokio

SoloJacobs:
WIP

### DIFF
--- a/v2/robotmk/Cargo.lock
+++ b/v2/robotmk/Cargo.lock
@@ -117,113 +117,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-channel"
-version = "1.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "81953c529336010edd6d8e358f886d9581267795c61b19475b71314bffa46d35"
-dependencies = [
- "concurrent-queue",
- "event-listener",
- "futures-core",
-]
-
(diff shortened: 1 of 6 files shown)


jherbel (approved):
Very nice, thanks!

jherbel (changes requested):
After trying this out and then debugging for a long time, I found out that we cannot use `std::hint::spin_loop` in combination with `tokio`. I cannot say what the underlying problem is. Using the two in combination leads to random deadlocks. Replacing `std::hint::spin_loop` with `tokio::task::yield_now().await` works.

SoloJacobs:
> After trying this out and then debugging for a long time, I found out that we cannot use `std::hint::spin_loop` in combination with `tokio`. I cannot say what the underlying problem is. Using the two in combination leads to random deadlocks. Replacing `std::hint::spin_loop` with `tokio::task::yield_now().await` works.

The explanation is probably this: As soon as the scheduler enters the loop, there is no way for it to exit it. Since the scheduler here is the tokio scheduler, it will only pause a task, if the task itself reports, that it can be paused. Pausing here here just means awaiting a future. In your implementation this works correctly, since now the scheduler gets the chance to resume other tasks.

I am not sure what you changed, maybe we should have a look together at this change, and what other ideas I have here.

jherbel (approved):

PR #434, opened 2023-11-09
Improve task termination

jherbel:
In addition to attempting to kill the process via its PID, we create a
flag file. Before starting the actual test run, the task checks if the
flag file still exists. If not, the task terminates.

This provides us with an additional mechanism for shutting down the task
in its early stage when the PID might not yet have been written to file.

### DIFF
--- a/v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
+++ b/v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
@@ -7,11 +7,10 @@ use robotmk::termination::{waited, Outcome, TerminationFlag};
 use anyhow::{bail, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use chrono::{Duration as ChronoDuration, Local};
-use log::{debug, error};
-use std::fs::{read_to_string, write};
+use log::{debug, error, warn};
+use std::fs::{read_to_string, remove_file, write};
 use std::process::Command;
 use std::str::FromStr;
-use std::thread::sleep;
 use std::time::Duration;
 use sysinfo::Pid;
 use tokio::task::yield_now;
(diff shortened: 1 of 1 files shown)


SoloJacobs (commented):
I'm unsure how this mechanism works. Where is the checking for the flag file done? (I spent 10min looking, but still unclear, sorry!)

v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
```diff
@@ -161,6 +163,14 @@ where
     ))
 }
 
+fn start_task(task_name: &str, path_run_flag: &Utf8Path) -> Result<()> {
+    debug!("Starting task {task_name}");
+    write(path_run_flag, "").context(format!("Failed to create run flag file {path_run_flag}"))?;
```
SoloJacobs, line 168: Does it really make sense to mix this with the other errors?

SoloJacobs (approved):
Hmm, I don't think this is the correct approach, but it's simple and works... so oti.

v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
jherbel: Not sure, but let's leave it like this for now. Hopefully we anyway find a better approach in the future.

jherbel:
> I'm unsure how this mechanism works. Where is the checking for the flag file done? (I spent 10min looking, but still unclear, sorry!)

The check happens in this line:
`format!("if not exist {} exit /b 1", paths.run_flag)`

This is one of the lines we write to batch file that is executed by the task. This line is placed after the line which writes the PID to file and before the line which executes the test run. When we want to cancel the task, we first delete the flag file and then we additionally try to retrieve the PID and terminate the process. In case the latter doesn't work, the task should still be in its initial phase (before starting the test run) and the removal of the file should make it terminate.

I fully agree that this approach is ugly and far from optimal, but I am out of ideas. I tested it and it seemed to to the job.

PR #438, opened 2023-11-13
Suite results: Include more data

jherbel:
The following data will be needed on the server side for computing check results:
* Timestamp when merged XML (rebot) was created. We explicitly don't use timestamps contained in the XML because these timestamps do not include timezone information, which can lead to issues if the Checkmk server is in a different timezone.
* execution interval
* timeout
* maximum number of attempts


racicLuka (approved):
LGTM

PR #440, opened 2023-11-14
remove TerminationFlag

SoloJacobs:
CMK-15137


jherbel (changes requested):
Let's also change the names of the corresponding variables to `cancellation_flag` or `cancellation_token` or so. Otherwise, we have things such as `termination_flag.is_cancelled()` where the naming is inconsistent (termination and cancellation are mixed).

PR #444, opened 2023-11-30
improved tests for partner day

simonmeggle:


racicLuka (approved):
As far as I know, LGTM

PR #445, opened 2023-12-01
Component test for scheduler

jherbel:


v2/robotmk/tests/test_scheduler.rs
```diff
@@ -0,0 +1,304 @@
...
+    let mut rcc_config_diag_command = Command::new(rcc_binary_path);
+    rcc_config_diag_command
+        .arg("configuration")
+        .arg("longpaths");
+    let stderr = String::from_utf8(rcc_config_diag_command.output().await?.stderr)?;
+    assert_eq!(stderr, "OK.\n");
+    Ok(())
+}
+
+fn directory_entries(directory: impl AsRef<Path>, max_depth: usize) -> Vec<String> {
```
SoloJacobs, line 288: This needs to be sorted, `WalkDir::sort_by`

v2/robotmk/tests/test_scheduler.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+
+async fn run_scheduler(test_dir: &Utf8Path, config: &Config, n_seconds_run: u64) -> Result<()> {
+    let config_path = test_dir.join("config.json");
+    write(&config_path, to_string(&config)?)?;
+    let run_flag_path = test_dir.join("run_flag");
+    write(&run_flag_path, "")?;
+
+    let mut robotmk_cmd = Command::new(cargo_bin("robotmk"));
```
SoloJacobs, line 140: I'm unsure about this. We could also expose library functions, which allow calling the scheduler directly.

v2/robotmk/tests/test_scheduler.rs
```diff
@@ -0,0 +1,304 @@
...
+use serde_json::to_string;
+use std::env::var;
+use std::ffi::OsStr;
+use std::fs::{create_dir_all, remove_file, write};
+use std::path::Path;
+use std::time::Duration;
+use tokio::{process::Command, time::timeout};
+use walkdir::WalkDir;
+
+#[tokio::test(flavor = "multi_thread")]
```
SoloJacobs, line 19: Do we really need a multithreaded test here?

.github/workflows/component_tests.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions-rust-lang/setup-rust-toolchain@v1.5.0
+        with:
+          target: x86_64-pc-windows-gnu
+      - run: cargo test --target=x86_64-pc-windows-gnu --test test_scheduler -- --nocapture
+        working-directory: ${{ github.workspace }}/v2/robotmk/
+        env:
+          TEST_DIR: C:\test_scheduler
+          RCC_BINARY_PATH: C:\artifact\windows64\rcc.exe
+          RUN_FOR: 240
+      - uses: actions/upload-artifact@v3
```
SoloJacobs, line 28: What is this for?

.github/workflows/component_tests.yaml
```diff
@@ -0,0 +1,30 @@
+---
+name: "Component tests"
```
SoloJacobs, line 2: System test seems more appropriate?

SoloJacobs (dismissed):
Somewhat ugly, but I guess it improves the current situation.

.github/workflows/component_tests.yaml
jherbel: Done

.github/workflows/component_tests.yaml
jherbel: This stores the results produced by the scheduler as an artifact. This is intended to facilitate debugging.

v2/robotmk/tests/test_scheduler.rs
SoloJacobs: Hmm, then we should use the windows-agent to run the scheduler, no?

SoloJacobs (approved):

v2/robotmk/tests/test_scheduler.rs
jherbel: Yes, we actually should, but that's more involved.,

PR #446, opened 2023-12-04
Improve error message in case of missing session

jherbel:
Before attempting to run a command in a user session via the task scheduler, we check if the user has an active session.

### DIFF
--- a/v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
+++ b/v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
@@ -46,6 +46,7 @@ pub fn run_task(task_spec: &TaskSpec) -> Result<RunOutcome> {
         "Running the following command as task {} for user {}:\n{}\n\nBase path: {}",
         task_spec.task_name, task_spec.user_name, task_spec.command_spec, task_spec.base_path
     );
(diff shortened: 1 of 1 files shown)


v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
```diff
@@ -89,6 +90,48 @@ impl From<&Utf8Path> for Paths {
...
+    ) {
+        return Ok(());
+    }
+    bail!("No active session for user {user_name} found")
+}
+
+fn check_if_user_has_active_session(user_name: &str, query_user_stdout: &str) -> bool {
+    for line in query_user_stdout.lines().skip(1) {
+        let words: Vec<&str> = line.split_whitespace().collect();
+        println!("{:?}", words);
```
SoloJacobs, line 115: Debugging?

SoloJacobs (approved):

SoloJacobs (approved):

v2/robotmk/src/bin/scheduler/sessions/schtasks.rs
jherbel: Done

PR #448, opened 2023-12-05
Dev/sol/spinloop

SoloJacobs:


SoloJacobs:
I decided that removing the spinloop in `start_run_flag_watch_thread` is too much, since windows can only monitor directories (rather than files).

Moreover, the scheduler behaviour is somewhat modified. See second commit message.

jherbel (approved):
Very nice, this is what I had hoped for :+1: 
We of course need to adjust the data model on the server side.

PR #449, opened 2023-12-07
Avoid running all suites at the same time at scheduling start

jherbel:
This could lead to a high system load. Instead, we start as soon as the current timestamp is fully divisible by the execution interval.

### DIFF
--- a/v2/robotmk/src/bin/scheduler/scheduling/scheduler.rs
+++ b/v2/robotmk/src/bin/scheduler/scheduling/scheduler.rs
@@ -4,10 +4,11 @@ use crate::internal_config::{GlobalConfig, Suite};
 use crate::logging::log_and_return_error;
 
 use anyhow::{bail, Result};
+use chrono::Utc;
 use log::error;
 use std::time::Duration;
 use tokio::task::{spawn_blocking, JoinSet};
-use tokio::time::interval;
+use tokio::time::{interval_at, Instant};
 use tokio_util::sync::CancellationToken;
 
 #[tokio::main]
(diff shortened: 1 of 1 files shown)


v2/robotmk/src/bin/scheduler/scheduling/scheduler.rs
```diff
@@ -47,7 +58,7 @@ async fn run_suite_scheduler(suite: Suite) {
 }
 
 async fn run_cleanup_job(cancellation_token: CancellationToken, suites: Vec<Suite>) {
-    let mut clock = interval(Duration::from_secs(300));
+    let mut clock = interval_at(compute_start_time(300), Duration::from_secs(300));
```
SoloJacobs, line 61: It seems to me that calling `tick().await` on `clock` has a very similiar effect.

SoloJacobs (commented):
Two general remarks about this change:
* It seems to me that calling `tick().await` on `clock` has a very similar effect. (at least if `Interval::now()` is relatively similar in all suites when the clocks are started.
* I think the change itself needs to clarify whether this is addressing a theoretical problem or whether there is something concrete, which we observed in practice.

v2/robotmk/src/bin/scheduler/scheduling/scheduler.rs
jherbel: Not sure what you mean. If we dont use `interval_at` but `interval`, the first tick will complete immediately.

jherbel:
@SoloJacobs I discussed this with Simon, this a real problem (though not a severe one). See for example here:
https://github.com/elabit/robotmk/actions/runs/7141059520/job/19447631131#step:5:345
Before, with `clokwerk`, this was not the case.

SoloJacobs (approved):

PR #454, opened 2023-12-12
Dev/sol/termination test ii

SoloJacobs:
Adds three commits to dev/sol/termination_test i

### DIFF
--- a/v2/robotmk/tests/test_scheduler.rs
+++ b/v2/robotmk/tests/test_scheduler.rs
@@ -17,6 +17,7 @@ use tokio::{process::Command, time::timeout};
 use walkdir::WalkDir;
 
 #[tokio::test]
+#[ignore]
 async fn test_scheduler() -> Result<()> {
     let test_dir = Utf8PathBuf::from(var("TEST_DIR")?);
     create_dir_all(&test_dir)?;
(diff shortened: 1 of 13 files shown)


jherbel (approved):

v2/robotmk/tests/timeout/lib/create_child.py
```diff
@@ -0,0 +1,6 @@
+import subprocess
+
+
+def spawn() -> None:
+    with subprocess.Popen(["python", "-c", "import time; time.sleep(10)"]) as p:
```
jherbel, line 5: We can directly use `time.sleep` without a subprocess

v2/robotmk/tests/test_scheduler.rs
```diff
@@ -17,6 +17,7 @@ use tokio::{process::Command, time::timeout};
 use walkdir::WalkDir;
 
 #[tokio::test]
+#[ignore]
```
jherbel, line 20: This currently leads to the test also being skipped by our system tests

jherbel (changes requested):

v2/robotmk/tests/timeout/lib/create_child.py
SoloJacobs: Done

v2/robotmk/tests/test_scheduler.rs
SoloJacobs: Done:
https://github.com/elabit/robotmk/actions/runs/7195151926/job/19597218491?pr=454

jherbel (approved):

PR #464, opened 2023-12-15
RCC environment building and rebot command: Run in specific session

jherbel:
Previously, we didn't necessarily run these two steps in the same session we executed the test run in, which might lead to issues.


racicLuka (approved):
LGTM

PR #469, opened 2023-12-21
Improve error handling

jherbel:


v2/robotmk/src/bin/scheduler/setup/rcc.rs
```diff
@@ -146,7 +144,7 @@ fn rcc_setup(global_config: &GlobalConfig, rcc_suites: Vec<Suite>) -> Result<Vec
 fn disable_rcc_telemetry(
     global_config: &GlobalConfig,
     suites: Vec<Suite>,
-) -> Result<(Vec<Suite>, Vec<Suite>)> {
+) -> Result<(Vec<Suite>, HashMap<String, String>), Cancelled> {
```
racicLuka, line 147: Not sure if I am correct, but shouldn't this be `AnyhowResult` here?

jherbel: No, this is correct. We previously used `AnyhowResult` (was called just `Result` before), however, the only error case was a termination:
```
RunOutcome::Terminated => bail!("Terminated"),
```
So now, it's much clearer what kind of error can happen. In general, we should use `AnyhowResult` as late as possible (whenever a function can encounters more than one type of error and wants to pass them along).

racicLuka: Aah, okay. Thanks for the clarification!

racicLuka (approved):

PR #470, opened 2024-01-04
Scheduler: Implement optional grace period

jherbel:
CMK-15492


jherbel:
> LGTM, one thing I don't understand: The ticket mentions GracePeriod runs after Phase 1. I guess this currently means `setup::general::setup`, which does not have a phase, since it takes very little time.

@SoloJacobs Yes, this is what it means. Also, we cannot report the general setup as a phase, since this step ensures that we can actually report results by creating the necessary directories.

PR #471, opened 2024-01-04
Scheduler: Delete Windows tasks also after successful completion

jherbel:
Previously, tasks were only deleted in the case of an error or if the task was cancelled. Also adds a test for this in the scheduler system test.


SoloJacobs (approved):
LGTM

PR #473, opened 2024-01-05
Update dependencies

jherbel:
For now, we need to pin `sysinfo` to 0.30.1, see the comment in Cargo.toml. This issue needs to be addressed separately.


SoloJacobs (approved):
Interesting :thinking:

PR #476, opened 2024-01-05
Rename binaries

jherbel:
* `robotmk_agent` --> `agent_plugin`
* `robotmk` --> `scheduler`


SoloJacobs (changes requested):
I think having `robotmk` in the binary names would be good

jherbel:
> I think having `robotmk` in the binary names would be good

Done

SoloJacobs (approved):

PR #481, opened 2024-01-11
Adjust permissions per user instead of for group "Users"

jherbel:
This group does not always exits, especially systems with non-English language settings are affected by this.


SoloJacobs (approved):

jherbel:
Had to resolve merge conflicts

PR #482, opened 2024-01-11
Relax check if user has active session when running task

jherbel:
We use the output of `query user` to check if there is a session. Unfortunately, this output is locale-dependent. Hence, checking if the session state is "Active" does not work on systems with a non-English locale. We therefore relax the check to asserting that there is a session at all and ignore the state for now. Hopefully, we find a better solution over time.


SoloJacobs (dismissed):

jherbel:
Had to resolve merge conflicts

PR #488, opened 2024-01-25
Dev/sol/mkp

SoloJacobs:
This pull request adds the mkp as binary to our repo. It is then exposed inside our release page as follows:
image


simonmeggle (approved):
LGTM

PR #492, opened 2024-02-17
vagrant: minor tweaks

SoloJacobs:
* The `checkmk_agent_protocol` variable was renamed in an update of the
 `checkmk.general` ansible collection.
* The chocolatey docs recommend against using `--force`, see https://docs.chocolatey.org/en-us/choco/commands/


racicLuka (approved):
Sorry, didn't see this previously

PR #497, opened 2024-03-04
time runtime suite executions

SoloJacobs:
Developers of synthetic tests are required to specify a timeout to ensure that their test does not run longer than specified. Previously, the ouput of the rebot command was used as proxy for this command. Currently, there are some bugs in the rebot command in v7. This causes the runtimes to be computed incorrectly.

We know time the runtime of each attempt instead.


racicLuka (approved):
LGTM

PR #500, opened 2024-03-06
Fix invalid escape sequence

SoloJacobs:
In Python 3 the character \ is used to escape unicode characters. Thus, the following syntax
```
"\\|"
```
is not valid. This is the corresponding documentation.
```
Changed in version 3.6: Unrecognized escape sequences produce a
DeprecationWarning. In a future Python version they will be a
SyntaxWarning and eventually a SyntaxError.
```

CMK-15744

### DIFF
--- a/checks/robotmk.py
+++ b/checks/robotmk.py
@@ -1190,7 +1190,7 @@ def lns(src, dest):
 def perfdata_label(name):
     """Removes unwanted characters from the instring.
     Replaces separators by underscores."""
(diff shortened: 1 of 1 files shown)


SoloJacobs:
I have read the CLA Document and I hereby sign the CLA

SoloJacobs:
recheck

simonmeggle (approved):
LGTM

PR #501, opened 2024-03-06
rmk: Add CoreMode to RCCProfileConfig enum

racicLuka:
Align the RCCProfileConfig enum with client side

CMK-16381


racicLuka:
This is my, probably wrong, attempt to alter the code.
The case where `global_config` matches `RCCProfileConfig::CoreMode` should never happen, but if it does, maybe this is the way to handle it?

SoloJacobs (dismissed):
LGTM. I would maybe say something about this being an internal error.

SoloJacobs (approved):

PR #508, opened 2024-03-12
rmk: Adjust scheduler for core mode (no-op)

racicLuka:
`RCCProfileConfig` is now optional. If the value is `None`, skip RCC profile configuration.

### DIFF
--- a/src/bin/scheduler/internal_config.rs
+++ b/src/bin/scheduler/internal_config.rs
@@ -176,10 +176,10 @@ mod tests {
                 results_directory: Utf8PathBuf::from("/results"),
                 rcc_config: RCCConfig {
                     binary_path: Utf8PathBuf::from("/bin/rcc"),
-                    profile_config: RCCProfileConfig::Custom(CustomRCCProfileConfig {
+                    profile_config: Some(RCCProfileConfig::Custom(CustomRCCProfileConfig {
                         name: "Robotmk".into(),
                         path: "/rcc_profile_robotmk.yaml".into(),
-                    }),
+                    })),
                 },
                 suite_groups: vec![
                     SequentialSuiteGroup {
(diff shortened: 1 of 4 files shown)


SoloJacobs:
I don't think this quite correct: `rcc_setup` and `setup` are still doing a lot of  setup. I think you should modify `src/bin/scheduler/rcc.rs` to not do any configuration of rcc and only return surviving_suites.

SoloJacobs (changes requested):
Please modify `setup` in src/bin/scheduler/setup/rcc.rs instead of `configure_curstom_rcc_profile`.

racicLuka:
So something like:
```
if global_config.rcc_config.profile_config.is_none() {
        sort_suites_by_grouping(&mut surviving_suites);
        Ok(surviving_suites);
    }
```
on line https://github.com/elabit/robotmk/blob/7f671ceec0965350b7be8a031356d7de15e27740/src/bin/scheduler/setup/rcc.rs#L26?

SoloJacobs:
Yes, but you can use pattern matching to unwrap  the profile and then it can be used in the body of the function. The rust tooling should be able to help you with this.

Also add some logging why that rcc configuration is skipped would be good.

In the end, I will write an integration test to check that rcc is never called.

racicLuka:
Sorry for the late response, I wasn't feeling well. 
I tried what you proposed, using pattern matching, but I can't get it to work. I just can't wrap my head around some concepts in Rust :/ 
If you have time, I would appreciate if we could take a look together. Thanks! 

PS. At least this way the profile configuration is skipped. I tested it in a site.

SoloJacobs (approved):
LGTM, let's wait for the test before we release this.

SoloJacobs (approved):

PR #509, opened 2024-03-18
improve log ouput for termination test

SoloJacobs:
It appears that sometimes the RobotFramework can spawn a huge number of process, which causes `print_process_tree` to never terminate. We do not have a reliable way to reproduce this, but at least we improve the test output.


racicLuka (approved):
LGTM

PR #510, opened 2024-03-19
Clearer distinction between core and enterprise mode

jherbel:


jherbel:
Hi @racicLuka  and @SoloJacobs, I though a bit about the distinction between core and enterprise mode. This might be a cleaner approach, but I am not 100% sure, so I would like to ask for your opinion. If we agree to go ahead with this, I will take polish it to make the CI happy.

jherbel:
@SoloJacobs und @racicLuka a simpler alternative might be: We revert Luka's last commit and instead add a simple check if there any RCC suites in `setup::rcc`. Something like
`    if rcc_suites.is_empty() {
        sort_suites_by_grouping(&mut surviving_suites);
        return Ok(surviving_suites);
    }
`
This would avoid doing any kind of RCC stuff in the core mode. Furthermore, we would also skip the RCC setup in enterprise mode if there are no suites using RCC. This is probably desirable, but we might want to briefly check with Simon.

racicLuka:
IMO, the second solution is, as you said, simpler and achieves what we want. Which makes it quite desirable.

On the other hand, I do like the first solution. It adds a clear distinction between the two modes, but I am not sure if it adds unnecessary code complexity. Maybe we don't need it at this point.

SoloJacobs:
I'm under the impression that I don't quite understand this issue, yet.
* Having the config represent that we are in CoreMode is definitely a good change, and I also think constructing a `Suite` object is appropriate.
* Are we actually solving an implementation issue in this commit? I was under the impression that functionality-wise the current state is correct. Currently, we only return surviving suites from the `rcc::setup` step. Is this somehow unsafe?

jherbel:
This change is only about solving an implementation issue. Functionality-wise, everything is fine. What I don't like about the current implementation is that the distinction between core and enterprise mode is effectively done via `Option<RCCProfileConfig>`, which seems strange. I currently see two (in my opinion) better ways:

1. Distinction at the external config level, which is what this change does.
2. Simply check if there are no RCC suites and skip the RCC setup in this case. This of course does not directly distinguish between the two modes, however, I don't think we necessarily have to distinguish atm.

So the questions are:

1. Do you agree with the my point that distinguishing via `Option<RCCProfileConfig>` is not optimal?
2. If so, which alternative solution  do you prefer (of course, other suggestions are welcome).

racicLuka:
1. Provided with this information, I do agree that distinguishing using `Option<RCCProfielConfig>` is not the best way. Moving it one level up to `Option<RCCConfig>` would be a better.
2. Even though it adds complexity to the code, I would prefer the first solution. It makes the distinguishment way clearer and stricter. Checking if the `rcc_suites` is empty does solve the problem, but it doesn't add any context.

If we agree on implementing the first solution and if you don't mind Joerg I would like to take over. It would be a nice way to practice some more Rust. At the end, I would assign you to review everything and give the final OK.

jherbel:
Using `Option<RCCConfig>` at the external config level won't work, because you need the RCC binary path to build the internal `Suite` structs from the external config. So if you want to distinguish at the external config level, you have to do essentially what I did in this PR. Otherwise, we allow invalid configurations (no RCC config but a suite wants to use RCC), which we really should avoid.

That's why I am a bit torn between the two options I mentioned above. @racicLuka and @SoloJacobs, what do you think?

SoloJacobs:
Whiteboard comment, but: 

There should probably be a top-level key, which distinguishes CoreConfig from EnterpriseConfig at the external config level. That way a mix of CoreMode with an RCC field is a configuration error. This seems to me to be the correct approach, since each field might potentially have an "RCC" version. If the subfields are the same/not specific to RCC, then they can code share the serde schema.

As we pass to the internal configuration level, we can have suites, which are with or without rcc binary. They don't to necessarily need to be specific to CoreMode, since this is enforced at configuration time.

It is not true that we don't need to distinguish between CoreMode and Enterprise: If the Scheduler has not been updated on the WindowsNode (ie has enterprise mode), but the CheckMK site has CoreMode, then the user is getting rcc for free until the bakery is triggered. '_check_licensing' must check what kind of configuration the Scheduler is using at the while checks are running. Thus, I think option two is inappropriate.

I'm fine with both approaches. Given that we have little time this sprint, I was going to leave things as they are. But I would be very happy if have aclean solution.

jherbel:
The requirement that we need to ensure on the server side that the scheduler and the server agree regarding licensing is new to me. But if that is the case, then we definitely need to do essentially what I did in this PR. @racicLuka do you have time to finish this PR (mainly the CI) and implement the corresponding changes on the server side? If so, I would suggest to go ahead with this. However, if there are more important tasks atm, then I would suggest to revert the last commit and instead implement skipping the RCC setup if there are no RCC suites, since this can be done quite fast.

SoloJacobs:
@jherbel Ah, I did not realize that Simon already decided against this requirement. I simply deduced that we have this requirement based on this comment from [CMK-16309](https://jira.lan.tribe29.com/browse/CMK-16309):

> [Luka Racic](https://jira.lan.tribe29.com/secure/ViewProfile.jspa?name=luka.racic) there might be an additional issue with the MKP: After installing it, the bakery rule page will show that RCC is not available for already configured rules. However, when baking agents, RCC will be configured if the user didn't change the rule in between. This could be solved in two ways:
> 
> 1) We tell our users that they have to run `cmk-update-config` after installing the MKP.
> 
> 2) We adapt the bakery plugin accordingly. For RCC (on the suite level and for the overall configuration), we could check if the core mode is active and simply ignore what the rule content says.
> 
> Please discuss this with [Simon Meggle](https://jira.lan.tribe29.com/secure/ViewProfile.jspa?name=simon.meggle) .

Luka and I agree, that solution one is still much more appropriate, and it should be done eventually. However, for the time being we can implement solution two. Then we should address the content of your comment, and then we start working on having an explicit external config.

@jherbel @racicLuka Please coordinate between the two of you, who does what.

jherbel:
Ah, ok. My comment addresses a slightly different, but of course related issue: After you install the MKP, the bakery rule page will display that you are in core mode and that RCC is disabled, also for already existing rules. This is because the corresponding transform is applied before rendering the rule. However, when baking agents, the transform is not applied, so you still get RCC baked in if you don't update your configuration (`cmk-update-config` or opening and saving the rule).

By implementing a check whether the license on the server side matches the license on the client side, we could make customers aware of a mismatch and force an agent update. However, the customer would still have to somehow update existing bakery rules right after installing the MKP before doing the update, otherwise, the baked agents will still be from the enterprise license and might contain RCC suites. So my comment in [CMK-16309](https://jira.lan.tribe29.com/browse/CMK-16309) targets avoiding this update step.

I would suggest to go ahead as follows: @racicLuka reverts the last commit and implements solution number two for the time being. Note that this includes a small change in behavior also for enterprise customers: If an enterprise customer has no RCC suites configured, no RCC setup will happen, which is desirable I would say. Then, once I am back, we discuss if solution number one is needed and to what extend (including any potential server-side enforcement). Since the enforcement is not part of the 2.3 requirements (at least not that I am aware of, I also briefly checked some tickets this morning to make sure), there is no problem in not having this enforcement in 2.3. If it turns out that we want it, we can always include it in a patch release.

racicLuka:
Sounds good. I will continue as agreed. Thanks for the input!

jherbel:
Ok, thanks for the fruitful discussion. I will close this PR for now, we can re-open it if necessary.

jherbel:
Ok, thanks for the fruitful discussion. I will close this PR for now, we can re-open it if necessary.

PR #520, opened 2024-04-16
Renaming: Suite --> Plan

jherbel:


src/results.rs
```diff
@@ -1,11 +1,11 @@
...
+use crate::config::PlanMetadata;
 use crate::section::{WritePiggybackSection, WriteSection};
 use camino::{Utf8Path, Utf8PathBuf};
 use serde::Serialize;
 use std::collections::HashMap;
 
-pub fn suite_results_directory(results_directory: &Utf8Path) -> Utf8PathBuf {
-    results_directory.join("suites")
+pub fn plan_results_directory(results_directory: &Utf8Path) -> Utf8PathBuf {
+    results_directory.join("plans")
```
SoloJacobs, line 8: This is probably an incompatible change. What are users expected to see/do?

src/results.rs
```diff
@@ -62,18 +62,18 @@ pub enum EnvironmentBuildStage {
...
     pub config: AttemptsConfig,
-    pub metadata: SuiteMetadata,
+    pub metadata: PlanMetadata,
 }
 
-impl WritePiggybackSection for SuiteExecutionReport {
+impl WritePiggybackSection for PlanExecutionReport {
     fn name() -> &'static str {
-        "robotmk_suite_execution_report"
+        "robotmk_plan_execution_report"
```
SoloJacobs, line 76: Another change, which is user facing.

src/suites.rs
```diff
@@ -78,17 +78,17 @@ fn run_attempt(
     cancellation_token: &CancellationToken,
     output_directory: &Utf8Path,
 ) -> Result<(AttemptOutcome, Option<Utf8PathBuf>), Cancelled> {
-    let log_message_start = format!("Suite {}, attempt {}", id, attempt.index);
+    let log_message_start = format!("Plan {}, attempt {}", id, attempt.index);
 
     let run_outcome = match session
         .run(&RunSpec {
-            id: &format!("robotmk_suite_{}_attempt_{}", id, attempt.index),
+            id: &format!("robotmk_plan_{}_attempt_{}", id, attempt.index),
```
SoloJacobs, line 85: What does this ID do again?

src/results.rs
jherbel: I will write a werk in Checkmk about this. Users will have uninstall the agent and delete the Robotmk output directory, that is probably the safest way. I think before the release we can still do this, afterwards, we would have to at least consider this in more detail.

src/results.rs
jherbel: I think this should be addressed with the comment above.

src/suites.rs
jherbel: This ID is used in the Windows task scheduler as the task ID

SoloJacobs (approved):

PR #522, opened 2024-04-17
Add suite name to `PlanMetadata`

jherbel:
CMK-17005


SoloJacobs:
I thought we agreed, that I don't have to participate in this change?

jherbel:
My understanding was that you didn't want to participate in any discussions, I though reviewing would be ok. But no problem, I'll add Luka.

racicLuka (approved):
LGTM

PR #540, opened 2024-05-23
Allow unzipping during setup

SoloJacobs:
CMK-17430

I guess it will take me a bit longer to get the permissions for the unzipped directories working. Can you maybe have a look at this draft, which only works for non-headed execution.

### DIFF
--- a/.github/workflows/system_tests.yaml
+++ b/.github/workflows/system_tests.yaml
@@ -13,6 +13,10 @@ jobs:
           name: rcc
           path: C:\
       - uses: actions/checkout@v4
+      - run: mkdir C:\managed_robots
+      - run: cd data\retry_rcc\
+      - run: tar -c -a -f C:\managed_robots\retry_rcc_defN.zip .
+      - run: cd ..\..
       - uses: actions-rust-lang/setup-rust-toolchain@v1.5.0
         with:
           target: x86_64-pc-windows-gnu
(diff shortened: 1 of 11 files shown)


SoloJacobs:
Some similiar ideas:
* The checkmk bakery also transfers packages (it uses an .msi instead). So, we could also build this functionality into the Checkmk server.
*` rcc robot wrap` is essentially the same as here, but sets some files to executable.

jherbel:
This looks quite promising already. There main point currently unclear to me is how to best handle the various paths. The bakery plugin currently builds eg. the path to the robot file from the base directory and the relative path specified by the user. However, the directory the scheduler unzips into is probably a different directory. So we could of course adjust the bakery plugin, but then we would have a bit of duplicated business logic, namely how to construct the path to the robot file for managed robots (basically `managed_dir/plan_id/...`). Maybe this is perfectly fine, but I think we should briefly discuss this. Ping me please.

.github/workflows/system_tests.yaml
```diff
@@ -13,6 +13,11 @@ jobs:
           name: rcc
           path: C:\
       - uses: actions/checkout@v4
+      - run: mkdir C:\managed_robots
+      - run: cd data\retry_rcc\
+      - run: tar -c -a -f C:\managed_robots\retry_rcc_defN.zip .
+      - run: dir data
```
jherbel, line 19: Can be removed

jherbel (changes requested):
Only one small comment, looks good otherwise :+1:

jherbel (approved):

PR #546, opened 2024-06-04
Assemble robot command line completely internally

jherbel:
Previously, the configuration contained both command-line arguments and variable/argument files (since the latter must be made absolute according to the plan source). Now, the configurations has explicit fields for each supported command-line argument.


src/bin/scheduler/internal_config.rs
```diff
@@ -266,7 +287,15 @@ mod tests {
             plans[0].robot,
             Robot {
                 robot_target: Utf8PathBuf::from("/synthetic_tests/rcc/tasks.robot"),
-                command_line_args: vec![],
+                top_level_suite_name: None,
```
SoloJacobs, line 290: I would prefer, if `Robot` would construct a `command_line_args`.

SoloJacobs (changes requested):

SoloJacobs (approved):

PR #552, opened 2024-06-16
Dev/sol/sec

SoloJacobs:


src/bin/scheduler/setup/rcc.rs
```diff
@@ -260,20 +262,20 @@ fn enable_long_path_support(
...
-    command_spec.add_arguments(["holotree", "init"]);
+    command_spec.add_arguments(["holotree", "init", "--revoke"]);
     let mut succesful_plans = vec![];
     let mut failed_plans: HashMap<String, String> = HashMap::new();
 
     for (session, plans) in plans_by_sessions(plans) {
         debug!("Running {} for `{}`", command_spec, &session);
         let session_id = &format!(
-            "holotree_initialization_{}",
+            "holotree_disabling_{}",
```
jherbel, line 278: let's also call it `"holotree_disabling_sharing_{}"`

tests/test_scheduler.rs
```diff
@@ -397,19 +396,24 @@ async fn assert_rcc_files_permissions(
...
-    assert!(stdout.contains("holotree-global-shared                ...  \"true\""));
+    let diagnostics = read_configuration_diagnostics(&rcc_config.binary_path)?;
+    assert_eq!(
+        diagnostics
+            .details
+            .get("telemetry-enabled")
+            .unwrap()
+            .as_str(),
+        "false"
+    );
```
jherbel, line 407: In the last commit, we should explicitly test that the shared holotree is disabled if possbile

src/rcc.rs
```diff
@@ -0,0 +1,242 @@
+use camino::Utf8Path;
```
jherbel, line 1: Since we only use this for testing, let's move it to the test folder

jherbel (changes requested):
Only minor comments, nice work!

src/bin/scheduler/setup/rcc.rs
SoloJacobs: Done

tests/test_scheduler.rs
SoloJacobs: Done

src/rcc.rs
SoloJacobs: Done

jherbel (approved):

PR #554, opened 2024-06-20
Clearly distinguish between relative and absolute paths in config

jherbel:


jherbel:
@SoloJacobs Not 100% sure about this, what is your opinion?

SoloJacobs:
I'm not sure. Building type safe abstractions over the filesystem seems a bit overkill for what we are doing.

We also have some other areas, which are more in need of attention.

Maybe ask Mathias? He is part of the elabit organization, and thus can review here on GH.

jherbel:
@Synss We are unsure whether this is overengineering. What is your opinion on this? We don't need a detailed review of course. Feel free to ping us for discussions.

Synss:
I agree with Solomon. I don't see how distinguishing between absolute and relative paths is relevant. If anything, it will make the code harder to test.

jherbel:
Ok, thanks for the feedback, then I'll close this PR.

PR #558, opened 2024-06-21
Improve system tests for managed robot

jherbel:
* Fix zip path and robot yaml path for managed robot
* Assert that unzipped robot directory has correct permissions
* Assert that managed robot produced results

### DIFF
--- a/.github/workflows/system_tests.yaml
+++ b/.github/workflows/system_tests.yaml
@@ -14,9 +14,7 @@ jobs:
           path: C:\
       - uses: actions/checkout@v4
       - run: mkdir C:\managed_robots
-      - run: cd data\retry_rcc\
-      - run: tar -c -a -f C:\managed_robots\retry_rcc_defN.zip .
-      - run: cd ..\..
+      - run: tar --create --auto-compress --directory tests\minimal_suite\ --file C:\managed_robots\minimal_suite_defN.zip *
       - uses: actions-rust-lang/setup-rust-toolchain@v1.9
         with:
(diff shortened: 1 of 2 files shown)


tests/test_scheduler.rs
```diff
@@ -157,7 +158,7 @@ fn create_config(
                     PlanConfig {
                         id: "managed_robot_zip".into(),
                         source: Source::Managed {
-                            zip_file: "C:\\zips\\restry_rcc_defN.zip".into(),
```
SoloJacobs, line 160: It probably makes sense to keep the type of the tar file in the file name, here defN. In the future we can then add other compression methods, and passwords, to know what does and doesn't work.

.github/workflows/system_tests.yaml
```diff
@@ -14,17 +14,15 @@ jobs:
           path: C:\
       - uses: actions/checkout@v4
       - run: mkdir C:\managed_robots
-      - run: cd data\retry_rcc\
-      - run: tar -c -a -f C:\managed_robots\retry_rcc_defN.zip .
-      - run: cd ..\..
+      - run: tar -c -a -C tests\minimal_suite\ -f C:\managed_robots\minimal_suite.zip *
```
SoloJacobs, line 17: Rule of Sven: In scripts, always use the long form of the arguments, i.e.,
-a --> --auto-compress
-c --> --create
-C --> --directory
-f --> --file

SoloJacobs (approved):
LGTM, with some OTIs

SoloJacobs (approved):

PR #561, opened 2024-06-24
Remove misleading comment in release workflow

jherbel:


.github/workflows/release.yaml
```diff
@@ -25,9 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [tests, system_tests, rcc, build_robotmk]
     steps:
-      - uses: actions/checkout@v4 # Determines the ref, which is used for the
```
SoloJacobs, line 28: How does the `git tag <tagname>` here work? Can we remove the action alltogether?

jherbel: actions/checkout@v4 does this:
> # The branch, tag or SHA to checkout. When checking out the repository that
> # triggered a workflow, this defaults to the reference or SHA for that event.
> # Otherwise, uses the default branch.

So this gives us exactly what we want, namely our robotmk repository with the commit from where the release was triggered checked out.

SoloJacobs: OK

SoloJacobs (approved):
LGTM

PR #564, opened 2024-06-25
Don't allow UserSession in Scheduler for Linux

SoloJacobs:


src/bin/scheduler/internal_config.rs
```diff
@@ -227,9 +227,7 @@ mod tests {
                 robot_yaml_path: Utf8PathBuf::from("robot.yaml"),
                 build_timeout: 300,
             }),
-            session_config: SessionConfig::SpecificUser(UserSessionConfig {
-                user_name: "user".into(),
-            }),
+            session_config: SessionConfig::Current,
```
jherbel, line 230: Could we make the tests platform-specific to not loose coverage?

jherbel (changes requested):

jherbel (approved):

PR #565, opened 2024-06-28
Replace icacls.exe with WinAPI

SoloJacobs:
CMK-17607

Some more testing revealed, that we don't actually need `GetNamedSecurityInfoW`. I am not sure why it is shown by Windows documentation (all links in commit). This cuts done the size of the change by half, which is already much better.

Also, it appears that `SetNamedSecurityInfoW` is used much more frequently than `icacls.exe`. So maybe we should reconsider adding this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,9 +205,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.100"
+version = "1.0.101"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


jherbel:
It doesn't look that bad anymore, I am leaning towards going ahead an replacing icacls.exe with this. What is your opinion?

SoloJacobs:
I have gone ahead and  implemented the change fully here: https://github.com/elabit/robotmk/pull/568. Let's discuss it there.

PR #568, opened 2024-07-08
Dev/sol/icacls replace

SoloJacobs:


SoloJacobs:
Positive points in favor of merging with +, against merging with -
* (+) More explicit dependencies supported by the language.
* (+) Fewer sub-process calls.
* (-) Users can easily manually fix/test the permission settings with `icacls.exe` command.
* (-) We get better error messages with `icacls.exe`
* (-) The `WinAPI` is different from `icacls.exe` in ways, which are not transparent to me. There are some arbitrary changes (see `test/test_scheduler.rs`. Moreover, I was mistaken in thinking `GetNamedSecurityInfoW` was redundant. Without this call, Windows will delete all permissions set previously.
* (-) Rust does not support sanitizers/heap inspection for Windows. I currently don't have a reliable to check whether the implementation is correct. Windows will terminate the program without a trace-back, if a double free occurs. Based on this pull request, such bugs are very easy to introduce.

In sum, I am against adding this change.

jherbel:
Thanks for the evaluation, I agree with your findings.

PR #569, opened 2024-07-11
Dev/sol/move

SoloJacobs:


src/bin/scheduler/setup/general.rs
```diff
@@ -19,25 +19,21 @@ pub fn setup(global_config: &GlobalConfig, plans: Vec<Plan>) -> AnyhowResult<Vec
...
-        setup_rcc_working_directories(&global_config.working_directory, surviving_plans);
+    let (surviving_plans, managed_failures) = setup_managed_directories(plans);
+    let (mut surviving_plans, working_dir_failures) =
+        setup_working_directories(global_config, surviving_plans);
     GeneralSetupFailures {
-        working_directory_permissions: plan_failures
+        working_directory_permissions: working_dir_failures
             .into_iter()
-            .chain(rcc_failures.into_iter())
+            .chain(managed_failures)
```
jherbel, line 36: We should report these errors in a separate, st. we can display a dedicated error message in checkmk. Otherwise, we would display "Adjusting working directory permissions for user configured to run this plan failed" also for these errors.

jherbel (changes requested):
Wow, nice work! Only a minor comment.

src/bin/scheduler/setup/general.rs
SoloJacobs: Done. We can consider just having just a `SetupFailures` section, where all failures have the current error + a message (for the summary). That's how Kubernetes does it. I think this could avoid this error and make testing the output easier.

jherbel: Yes, that's true. For now, let's do this:
```
#[derive(Default, Serialize)]
pub struct GeneralSetupFailures {
    pub working_directories: HashMap<String, String>,
    pub managed_directories: HashMap<String, String>,
}
```

Both fields can contain errors for directory creation or permission adjustment. We can specify which of the two it is in the message. In Checkmk, we can then report something like "Setting up working/managed directory failed".

jherbel (changes requested):
Almost there, just a minor change regarding the error reporting

src/bin/scheduler/setup/general.rs
SoloJacobs: Isn't that what I did? Or is it just about the naming?

jherbel: It's about the naming and about adding an additional hint to the error messages (directory creation/permissions)

jherbel (approved):

PR #572, opened 2024-07-18
Add --version to scheduler & agent plugin and use in release workflow

jherbel:


jherbel:
Release workflow was tested here: https://github.com/elabit/robotmk/actions/runs/9995013229
(created tag and pre-release have already been deleted)

.github/workflows/release.yaml
```diff
@@ -27,6 +18,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions-rust-lang/setup-rust-toolchain@v1.9
+        with:
+          target: x86_64-pc-windows-gnu
```
SoloJacobs, line 23: Why windows?

jherbel: Because of shameless copy and paste. Once we include linux in our releases, we anyway won't need to build the scheduler in this step again just to get the version. We will simply use the binary built for linux instead.

SoloJacobs (approved):

PR #574, opened 2024-07-23
Dev/sol/interrupt

SoloJacobs:
Not complete, but it makes sense to review the changes up to this point.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -23,6 +23,7 @@ jobs:
         with:
           components: rustfmt, clippy
           target: ${{ matrix.type.target }}
+      - run: echo "RUSTFLAGS=--cfg tokio_unstable" >> "$GITHUB_ENV"
       - run: cargo fmt -- --check
       - run: cargo test --all-targets --target ${{ matrix.type.target }} -- --skip test_scheduler
       - run: cargo run --example termination --target ${{ matrix.type.target }}
--- a/src/bin/scheduler/setup/general.rs
+++ b/src/bin/scheduler/setup/general.rs
@@ -284,7 +284,7 @@ fn clean_up_results_directory(global_config: &GlobalConfig, plans: &[Plan]) -> A
         &plan_results_directory(&global_config.results_directory),
         plans,
     )?;
-    results_directory_lock.release()
+    Ok(results_directory_lock.release()?)
 }
 
 fn clean_up_plan_results_directory(
--- a/src/bin/scheduler/setup/general.rs
+++ b/src/bin/scheduler/setup/general.rs
@@ -7,26 +7,47 @@ use anyhow::{anyhow, Context, Result as AnyhowResult};
 use camino::{Utf8Path, Utf8PathBuf};
 use log::error;
 use robotmk::environment::Environment;
+use robotmk::fs::{create_dir_all, remove_dir_all, remove_file};
+use robotmk::lock::LockerError;
 use robotmk::results::{plan_results_directory, SetupFailure};
 use std::collections::HashSet;
-use std::fs::{create_dir_all, remove_dir_all, remove_file};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum SetupError {
+    #[error("Failed to `{0}`")]
+    Unrecoverable(String),
+    #[error("Terminated")]
+    Cancelled,
+}
+
+impl From<LockerError> for SetupError {
+    fn from(value: LockerError) -> Self {
+        match value {
+            LockerError::Cancelled => Self::Cancelled,
+            value => Self::Unrecoverable(format!("{:?}", value)),
+        }
+    }
+}
+
+impl From<anyhow::Error> for SetupError {
+    fn from(value: anyhow::Error) -> Self {
+        Self::Unrecoverable(format!("{:?}", value))
+    }
+}
 
 pub fn setup(
     global_config: &GlobalConfig,
     plans: Vec<Plan>,
-) -> AnyhowResult<(Vec<Plan>, Vec<SetupFailure>)> {
+) -> Result<(Vec<Plan>, Vec<SetupFailure>), SetupError> {
     if global_config.working_directory.exists() {
-        remove_dir_all(&global_config.working_directory)
-            .context("Failed to remove working directory")?;
+        remove_dir_all(&global_config.working_directory)?;
     }
-    create_dir_all(&global_config.working_directory)
-        .context("Failed to create working directory")?;
+    create_dir_all(&global_config.working_directory)?;
     if global_config.managed_directory.exists() {
-        remove_dir_all(&global_config.managed_directory)
-            .context("Failed to remove managed directory")?;
+        remove_dir_all(&global_config.managed_directory)?;
     }
-    create_dir_all(&global_config.managed_directory)
-        .context("Failed to create managed directory")?;
+    create_dir_all(&global_config.managed_directory)?;
     setup_results_directories(global_config, &plans)?;
 
     let (surviving_plans, managed_dir_failures) = setup_managed_directories(plans);
@@ -214,10 +235,8 @@ fn setup_with_one_directory_per_user(
 }
 
 fn setup_results_directories(global_config: &GlobalConfig, plans: &[Plan]) -> AnyhowResult<()> {
-    create_dir_all(&global_config.results_directory)
-        .context("Failed to create results directory")?;
-    create_dir_all(plan_results_directory(&global_config.results_directory))
-        .context("Failed to create plan results directory")?;
+    create_dir_all(&global_config.results_directory)?;
+    create_dir_all(plan_results_directory(&global_config.results_directory))?;
     clean_up_results_directory(global_config, plans).context("Failed to clean up results directory")
 }
 
@@ -273,7 +292,10 @@ fn setup_managed_directories(plans: Vec<Plan>) -> (Vec<Plan>, Vec<SetupFailure>)
     (surviving_plans, failures)
 }
 
-fn clean_up_results_directory(global_config: &GlobalConfig, plans: &[Plan]) -> AnyhowResult<()> {
+fn clean_up_results_directory(
+    global_config: &GlobalConfig,
+    plans: &[Plan],
+) -> Result<(), SetupError> {
     let results_directory_lock = global_config
         .results_directory_locker
         .wait_for_write_lock()?;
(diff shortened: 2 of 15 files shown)


src/child_process_supervisor.rs
```diff
@@ -74,11 +76,37 @@ async fn wait_for_child(
...
+            {
+                use tokio::time::sleep;
+                interrupt(child);
+                tokio::select! {
+                    _ = child.wait() => { },
+                    _ = sleep(Duration::from_secs(10)) => {
+                        kill_child_tree(child);
+                        let _ = child.wait().await;
+                    },
+                };
```
jherbel, line 91: Would it make sense to also move this into `interrupt`?

jherbel (changes requested):
Nice, only one remark/question :+1:

src/child_process_supervisor.rs
SoloJacobs: Done

src/bin/scheduler/setup/general.rs
```diff
@@ -273,7 +298,10 @@ fn setup_managed_directories(plans: Vec<Plan>) -> (Vec<Plan>, Vec<SetupFailure>)
...
 }
 
-fn clean_up_results_directory(global_config: &GlobalConfig, plans: &[Plan]) -> AnyhowResult<()> {
+fn clean_up_results_directory(
+    global_config: &GlobalConfig,
+    plans: &[Plan],
+) -> Result<(), SetupError> {
     let results_directory_lock = global_config
         .results_directory_locker
         .wait_for_write_lock()?;
```
jherbel, line 307: I am surprised that this works as is. Apparently, the conversions implemented above via the `From` trait are applied automatically eg. here, right? Or am I misunderstanding?

src/fs.rs
```diff
@@ -0,0 +1,26 @@
+use camino::{Utf8Path, Utf8PathBuf};
+use std::fs;
+use std::io;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
```
jherbel, line 6: This seems unnecessary, since we have `impl From<anyhow::Error> for SetupError`. Our functions below could simply use `anyhow::Error` instead, since we will never dispatch on the variants of this enum anyway.

src/section.rs
```diff
@@ -69,13 +98,13 @@ pub trait WritePiggybackSection {
...
 
-fn read_entry(entry: Result<DirEntry, Error>) -> AnyhowResult<Section> {
+fn read_entry(entry: Result<DirEntry, walkdir::Error>) -> AnyhowResult<Section> {
     let entry = entry?;
     let raw = read_to_string(entry.path())?;
     Ok(serde_json::from_str(&raw)?)
 }
 
-pub fn read(directory: impl AsRef<Path>, locker: &Locker) -> AnyhowResult<Vec<Section>> {
+pub fn read(directory: impl AsRef<Path>, locker: &Locker) -> Result<Vec<Section>, WriteError> {
```
jherbel, line 107: Wouldn't `Result<Vec<Section>, LockerError>` be sufficient here?

jherbel (changes requested):
This is impressive work! For my understanding: For example, the commit
"Remove exit-code if shutting down during GeneralSetup"
does not actually remove the exit code on its own, right? Only in combination with the changes to main.rs in "Remove exit-code 1 if shutting down during write" the exit code is removed, right?

.github/workflows/tests.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
         with:
           components: rustfmt, clippy
           target: ${{ matrix.type.target }}
+      - run: echo "RUSTFLAGS=--cfg tokio_unstable" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
SoloJacobs: From my testing we do. I believe this is to the way the `cargo` command searches directories for this file, but I did not investigate this yet.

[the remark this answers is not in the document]
SoloJacobs: By this file, I mean `.cargo/config.toml`

src/bin/scheduler/setup/general.rs
SoloJacobs: Yes, the question mark operator is syntatic sugar. Here is a quote:
https://doc.rust-lang.org/reference/expressions/operator-expr.html

> When applied to values of the Result<T, E> type, it propagates errors. If the value is Err(e), then it will return Err(From::from(e)) from the enclosing function or closure. If applied to Ok(x), then it will unwrap the value to evaluate to x.

src/fs.rs
SoloJacobs: Yes, true. l like this style, since it does not require us to add the necessary context about the `path` at the call site of the function. I makes the control flow a bit more visible in my opinion, and also reduces duplication.

src/section.rs
SoloJacobs: The question mark operator only calls the `From` trait once. However, my implementation requires two `from` calls to turn Locker Errors into SetupErrors, and only one to convert a WriteError. Thus, in the setup functions working with the WriteError is easier.

.github/workflows/tests.yaml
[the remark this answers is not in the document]
jherbel: Ok

src/fs.rs
jherbel: What I meant is this:
```
pub fn create_dir_all<P: AsRef<Utf8Path>>(path: P) -> anyhow::Result<()> {
    fs::create_dir_all(path.as_ref()).context(...)
}
```
Wouldn't this be sufficient?

src/section.rs
jherbel: It seems strange that a function called `read` returns a `WriteError`. If possible, please us `LockerError` instead.

src/fs.rs
jherbel: What I meant is this:
```
pub fn create_dir_all<P: AsRef<Utf8Path>>(path: P) -> anyhow::Result<()> {
    fs::create_dir_all(path.as_ref()).context(...)
}
```
Wouldn't this be sufficient?

jherbel:
For my understanding: For example, the commit
"Remove exit-code if shutting down during GeneralSetup"
does not actually remove the exit code on its own, right? Only in combination with the changes to main.rs in "Remove exit-code 1 if shutting down during write" the exit code is removed, right?

jherbel (changes requested):

src/section.rs
SoloJacobs: Done

SoloJacobs:
@RobWalt I hope you don't mind me asking a question. Don't feel obliged to answer. 

We realized, that terminating our program with SIGINT/systemd would always result in a failed state for systemd. For this reason there are usually three outcomes for each function: `Ok(ComputationOutcome)`, `Err(TeminateWithError)`, `Err(TerminateWithoutError)`. 

This is reflected in the changes in this pull request.  However, every function in our code now has it's own `TerminateWithoutError` type, since `enum` variants cannot be shared between different `enum`s. Alternatively, we could have every function have the same `Error` types, but that means some functions will have errors in their return signature, which they don't actually produce. (This is the same as the `std` library chooses with `std::io::Error`).

With pythons duck typing this can been solved differently:
```
class TerminateWithError: pass

class TerminateWithoutError: pass

def setup() -> TerminateWithError | TerminateWithoutError | Plans: 
   ...
   
def rcc_setup() -> TerminateWithError |TerminateWithoutError | RccPlans:
   ...
```

How does one approach such a problem idiomatically in rust?

RobWalt:
@SoloJacobs no worries!

I'm not entirely sure what exactly your pain points are. From looking at your code, you have a lot of repetitive error types now. What came to mind was that you might be able to construct a generic error of the same form and implement the chore-y parts only once. E.g. something like

<details>

<summary>Code</summary>

```rust
use std::marker::PhantomData;

use thiserror::Error;

#[derive(Debug)]
pub struct Write;
#[derive(Debug)]
pub struct Build;
#[derive(Debug)]
pub struct Run;
#[derive(Debug)]
pub struct Lock;

// type WriteError = GenericError<Write, Unrecoverable>;
// type BuildError = GenericError<Build, Unrecoverable>;
// type RunError = GenericError<Run, Unrecoverable>;
type LockError = GenericError<Lock, LockVariants>;

fn main() {
    let e = LockError::new_other(LockVariants::Foo(String::from("Hello, World!")));
    println!("{e}");
    let e = e.cast_error::<Write, Unrecoverable>();
    println!("{e}");
    let e = e.cast_error::<Build, Unrecoverable>();
    println!("{e}");
    let e = e.cast_error::<Run, Unrecoverable>();
    println!("{e}");
}

#[derive(Error, Debug)]
pub enum ErrorVariant<E: std::error::Error> {
    #[error(transparent)]
    Other(E),
    #[error("Terminated")]
    Cancelled,
}

#[derive(Error, Debug)]
#[error("Failed to `{0}`")]
pub struct Unrecoverable(String);

#[derive(Error, Debug)]
pub enum LockVariants {
    #[error("Failed to open `{0}`")]
    Foo(String),
    #[error("Could not complete task")]
    Bar(String),
}

impl From<LockVariants> for Unrecoverable {
    fn from(value: LockVariants) -> Self {
        Self(format!("{value:?}"))
    }
}

#[derive(Error, Debug)]
#[error("{variant}")]
pub struct GenericError<Marker, NonCancelled: std::error::Error> {
    variant: ErrorVariant<NonCancelled>,
    _pd: PhantomData<Marker>,
}

impl<A, E: std::error::Error> GenericError<A, E> {
    pub fn new_other(variant: E) -> Self {
        Self {
            variant: ErrorVariant::Other(variant),
            _pd: PhantomData,
        }
    }
    pub fn new_cancelled() -> Self {
        Self {
            variant: ErrorVariant::Cancelled,
            _pd: PhantomData,
        }
    }
}

// instead of `From`, since impl `From` is impossible due to existing impls in the std lib
impl<A, EA: std::error::Error> GenericError<A, EA> {
    pub fn cast_error<B, EB: std::error::Error + std::convert::From<EA>>(
        self,
    ) -> GenericError<B, EB> {
        let variant = match self.variant {
            ErrorVariant::Other(v) => ErrorVariant::Other(EB::from(v)),
            ErrorVariant::Cancelled => ErrorVariant::Cancelled,
        };
        GenericError {
            variant,
            _pd: PhantomData,
        }
    }
}
```

</details>

Other than that there are ideas to use traits similar to `std::error::Error`, which you already mentioned. Maybe it's worth taking a look at how big ecosystem libraries handle that problem as well. I'd guess that especially networking crates like [`hyper`](https://docs.rs/hyper/latest/hyper/) would have to solve something similar due to the variety in HTTP response codes.

RobWalt:
@SoloJacobs Glad to hear that it helped! :) 🎉

PR #577, opened 2024-08-01
Dev/sol/interrupt

SoloJacobs:


examples/termination/conda.yaml
```diff
@@ -0,0 +1,10 @@
+# https://github.com/robocorp/rcc/blob/master/docs/recipes.md#what-is-in-condayaml
+
+channels:
+  - conda-forge
+
+dependencies:
+  - python=3.9.12               # Install obscure python version
+  - pip=22.1.2                  # Needed for installing rpaframework
+  - pip:
+      - rpaframework==22.5.2    # Needed for `robot` command
```
jherbel, line 10: Let's use robotframework directly, since this is faster to set up

examples/termination/conda.yaml
```diff
@@ -0,0 +1,10 @@
+# https://github.com/robocorp/rcc/blob/master/docs/recipes.md#what-is-in-condayaml
+
+channels:
+  - conda-forge
+
+dependencies:
+  - python=3.9.12               # Install obscure python version
```
jherbel, line 7: It doesn't matter too much, but I would suggest to go with the latest versions here

.github/workflows/tests.yaml
```diff
@@ -29,3 +29,4 @@ jobs:
       - run: cargo run --example termination --target ${{ matrix.type.target }}
       - run: cargo clippy --all-targets --target ${{ matrix.type.target }} -- --deny warnings
       - run: cargo build --target ${{ matrix.type.target }}
+      - run: cargo run --example termination
```
jherbel, line 32: We already have this a couple of lines above. We should also move this to the system tests for consistency, but let's do this later in a later change.

.github/workflows/system_tests.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
           RCC_BINARY_PATH: C:\windows64\rcc.exe
           RUN_FOR: 300
       - uses: actions/upload-artifact@v4
+        if: success() || failure()
```
jherbel, line 27: Can we use `if: always()` here?

jherbel (changes requested):

.github/workflows/tests.yaml
SoloJacobs: Done

examples/termination/conda.yaml
SoloJacobs: Done

examples/termination/conda.yaml
SoloJacobs: Done

.github/workflows/system_tests.yaml
SoloJacobs: `if: always()` will cause this change to also trigger, if the workflow was cancelled, so does not make much sense here.

jherbel (dismissed):

jherbel (approved):

PR #579, opened 2024-08-05
Miscellaneous improvements

jherbel:


src/section.rs
```diff
@@ -91,7 +91,7 @@ pub trait WritePiggybackSection {
     {
         let section = Section {
             name: Self::name().into(),
-            content: serde_json::to_string(&self).unwrap(),
```
SoloJacobs, line 94: I believe this should actually stay an unwrap, since it the function should only error, if serde_json was used incorrectly. Thus, the error needs to stay unrecoverable (but maybe a bit more context could be good).

src/child_process_supervisor.rs
```diff
@@ -108,13 +108,26 @@ fn kill_child_tree(child: &tokio::process::Child) {
...
-        let gid = getpgid(Some(Pid::from_raw(pid as i32))).unwrap();
-        killpg(gid, Signal::SIGKILL).unwrap();
+        let gid = match getpgid(Some(Pid::from_raw(pid as i32))) {
+            Ok(gid) => gid,
+            Err(error) => {
+                error!(
+                    "Failed to retrieve process group ID of process {pid}, 
+                     cannot proceed with interruption. Error message:\n{error:#?}"
+                );
+                return;
```
SoloJacobs, line 124: Is the early return here really correct? Killing the tree seems more consistent

SoloJacobs (changes requested):

src/section.rs
jherbel: Done

src/child_process_supervisor.rs
jherbel: Done

SoloJacobs (approved):

SoloJacobs:
Merging this, so I can rebase :)

PR #606, opened 2024-09-19
Dev/sol/remote origin

SoloJacobs:


src/command_spec.rs
```diff
@@ -93,6 +108,7 @@ mod tests {
                         String::from("--option"),
                         String::from("value"),
                     ],
+                    envs: vec![],
```
jherbel, line 111: Can we test environment variables here? Or are they not part of the formatted string?

src/command_spec.rs
```diff
@@ -1,11 +1,13 @@
 use std::convert::From;
+use std::ffi::OsString;
 use std::fmt::{Display, Formatter, Result};
 use std::process::Command;
 
 #[derive(Debug, PartialEq)]
 pub struct CommandSpec {
     pub executable: String,
     pub arguments: Vec<String>,
+    envs: Vec<(String, String)>,
```
jherbel, line 10: (Sry, didn't know where else to put this, since I can only comment on changed lines)

`impl From<&CommandSpec> for tokio::process::Command` also needs to take `envs` into account. `tokio::process::Command` implements `From<std::Process::Command>`, so we let's just use that.

jherbel (changes requested):

src/command_spec.rs
SoloJacobs: Ugh, it does work, but that's probably not what we want, if we ever add any secrets to our environment variables.

SoloJacobs: Only works on Linux, oddly enough. Added the test for documentation.

src/tasks/windows.rs
```diff
@@ -325,10 +339,11 @@ mod tests {
                 &Paths::from(Utf8PathBuf::from("C:\\working\\plans\\my_plan\\123\\0").as_ref())
             ),
             "@echo off
+setlocal
 echo Robotmk: running task robotmk_task. Please do not close this window.
-\"C:\\\\somewhere\\\\rcc.exe\" \"mandatory\" \"--some-flag\" \"--some-option\" \"some-value\" \
-> C:\\working\\plans\\my_plan\\123\\0.stdout 2> C:\\working\\plans\\my_plan\\123\\0.stderr
-echo %errorlevel% > C:\\working\\plans\\my_plan\\123\\0.exit_code"
+
```
jherbel, line 344: Let's include one environment variable in the test

src/command_spec.rs
```diff
@@ -51,6 +64,17 @@ impl CommandSpec {
             .extend(arguments.into_iter().map(|s| s.as_ref().into()));
         self
     }
+
+    pub fn add_env(&mut self, key: String, value: String) -> &mut Self {
```
jherbel, line 68: Let's add a test for this below

src/command_spec.rs
```diff
@@ -93,6 +117,7 @@ mod tests {
                         String::from("--option"),
                         String::from("value"),
                     ],
+                    envs: vec![],
```
jherbel, line 120: Could we also test for environment variables here? Not by using `format!` but instead by using `Command::{get_program, get_args, get_envs}`.

jherbel (changes requested):

jherbel (approved):

PR #608, opened 2024-09-23
Don't allow Microsoft telemetry

SoloJacobs:
It is unclear why MSVC telemetry is running in the background, eventhough we are using the GNU toolchain. However, if don't turn it off, then we have flakes in `examples/termination/main.rs`.

### DIFF
--- a/.github/workflows/system_tests.yaml
+++ b/.github/workflows/system_tests.yaml
@@ -27,6 +27,12 @@ jobs:
           name: rcc
           path: C:\
 
+      # MSVC uses vctip.exe for telemetry. vctip.exe is started as a child of termination.exe. This
+      # can cause CI failures, if "vctip.exe" does not terminate before `get_children` is called.
+      # It is unclear why MSVC is running, despite target=x86_64-pc-windows-gnu. The following
+      # command is intended turn off telemetry via vctip.exe.
+      - shell: pwsh
+        run:  Get-ChildItem -Filter vctip.exe -Recurse "C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC" | rm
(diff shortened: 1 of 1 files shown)


.github/workflows/system_tests.yaml
```diff
@@ -27,6 +27,10 @@ jobs:
           name: rcc
           path: C:\
 
+      # We need to opt-out of telemetry. Not sure why this is started though, since we use GNU for
+      # compilation, not MSVC. Telemetry will interfere with `examples/termination/main.rs`.
+      - shell: pwsh
+        run:  Get-ChildItem -Filter vctip.exe -Recurse "C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC\\Tools\\MSVC" | rm
```
jherbel, line 33: Is the problem that vctip.exe is started at some point which messes up our process tree?

SoloJacobs: Yes, it was probably always the case, but it failed yesterday, after I updated the cargo.lock: https://github.com/elabit/robotmk/actions/runs/10996567682/attempts/1

jherbel: Hrmpf. Ok, then let's hope it's never started before we remove it. But I guess we just have to try out and see. Please expand the comment above a bit to explain that apparently, vctip.exe might be started randomly at some point during the test run, messing up the process tree.

jherbel (changes requested):

.github/workflows/system_tests.yaml
SoloJacobs: I think it is started by the `cargo --examples termination` invocation, otherwise it would not interfere with our test. I added the information I have.

jherbel (approved):

PR #609, opened 2024-09-25
FIX elapsed of nodes 

m3rlinux:
The XML structure doesn't have the attributes starttime and endtime for status nodes. Instead, it already has elapsed


m3rlinux:
I have read the CLA Document and I hereby sign the CLA

m3rlinux:
recheck

PR #610, opened 2024-09-25
Update robotmk.py

m3rlinux:
Using the latest version of Robot Framework, the XML structure no longer has the attributes 'starttime' and 'endtime' for each status node. Instead, it now includes the attribute 'elapsed.' With this minor change, the plugin can be used with the latest version of Robot Framework as well.

### DIFF
--- a/checks/robotmk.py
+++ b/checks/robotmk.py
@@ -553,15 +553,9 @@ class RobotItem(object):
         return subnodes
 
     def _get_node_elapsed_time(self):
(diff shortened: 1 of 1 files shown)


SoloJacobs:
@simonmeggle I think you need to adjust the permissions for the environment variable to include the Personal Access Token (or we turn off the bot for v1/dev, since we did not really write the bot with your code in mind).
image
Review is also up you. If I recall correctly the `@elapsed` field served incorrect data in some versions of Robot framework.

PR #612, opened 2024-09-30
Dev/sol/hololib impl

SoloJacobs:


src/bin/scheduler/build.rs
```diff
@@ -63,30 +63,82 @@ fn build_environment(
...
+                return Err(Cancelled {});
+            }
+            Err(e) => {
+                let log_error = e.context(anyhow!(
+                    "Environment import failed, plan {id} will be dropped. See {} for stdio logs",
+                    base_path,
+                ));
+                error!("{log_error:?}");
+                return Ok(BuildOutcome::Error(format!("{log_error:?}")));
+            }
```
jherbel, line 124: Would it make sense to outsource this into a separate function to reduce code duplication?

jherbel (changes requested):

src/bin/scheduler/build.rs
SoloJacobs: I think the similiarity here is just coincidental. For example, if ever need to handle the exit codes, then this would need to be de-duplicated. I usually like to:
* only abstract once there are three or more examples,
* don't code share messaging/ui elements (in this case the log messages).

We can deviate from my gut feeling here, but the duplication here was deliberate. Adapting duplicated code is much easier, then adapting code with unnecessary coupling.

jherbel: Sounds reasonable to me, thanks for clarifying.

jherbel (approved):

PR #626, opened 2024-11-04
Set directory ownership explicitly during setup

jherbel:


src/bin/scheduler/setup/directories.rs
```diff
@@ -368,6 +370,66 @@ fn clean_up_results_directory(
...
+    #[cfg(unix)]
+    user_id: u32,
+    #[cfg(unix)]
+    group_id: u32,
+    #[cfg(windows)]
+    user: String,
+}
+
+impl OwnershipSetter {
+    pub fn make_for_current_user() -> AnyhowResult<Self> {
```
SoloJacobs, line 414: This result type seems unnecessary?

src/bin/scheduler/setup/directories.rs
```diff
@@ -368,6 +370,66 @@ fn clean_up_results_directory(
...
+        }
+        #[cfg(windows)]
+        {
+            use super::windows_permissions::set_ownership;
+            set_ownership(&self.user, target, recursive)
+        }
+    }
+
+    #[cfg(unix)]
+    fn take_ownership_recursive_unix(&self, target: &Utf8Path) -> std::io::Result<()> {
```
SoloJacobs, line 451: Does handle symlinks correctly? What if there is a cycle?

src/bin/scheduler/setup/directories.rs
```diff
@@ -368,6 +370,66 @@ fn clean_up_results_directory(
...
+        }
+        #[cfg(windows)]
+        {
+            Ok(Self {
+                user: whoami::username(),
+            })
+        }
+    }
+
+    pub fn take_ownership(&self, target: &Utf8Path, recursive: bool) -> AnyhowResult<()> {
```
SoloJacobs, line 430: Having two functions here without the bool argument would be cleaner. Maybe the recursive function can call the non-recursive one (to deduplicate the `format!` string)

src/bin/scheduler/setup/directories.rs
SoloJacobs, line 451: Dont include `unix` in the function name. This information will always be clear to the reader anyway.

src/bin/scheduler/setup/directories.rs
```diff
@@ -368,6 +370,66 @@ fn clean_up_results_directory(
...
+        {
+            Ok(Self {
+                user_id: unsafe { libc::getuid() },
+                group_id: unsafe { libc::getgid() },
+            })
+        }
+        #[cfg(windows)]
+        {
+            Ok(Self {
+                user: whoami::username(),
```
SoloJacobs, line 425: I think this output needs to be sanitized? `whoami` crate can return `Unknown` in some circumstances? Do we really want to change the permissions of the directories to that of the user `Unknown`? I feel that the whoami crate is creating more problems then is solving tbh.

src/bin/scheduler/setup/windows_permissions.rs
```diff
@@ -94,6 +94,18 @@ pub fn reset_access(target_path: &Utf8Path) -> anyhow::Result<()> {
     })
 }
 
+pub fn set_ownership(user: &str, target_path: &Utf8Path, recursive: bool) -> anyhow::Result<()> {
```
SoloJacobs, line 97: Why not use the same signature as for linux?

src/bin/scheduler/setup/directories.rs
```diff
@@ -20,6 +18,10 @@ pub fn setup(
     global_config: &GlobalConfig,
     plans: Vec<Plan>,
 ) -> Result<(Vec<Plan>, Vec<SetupFailure>), Terminate> {
+    let ownership_setter = OwnershipSetter::make_for_current_user()?;
```
SoloJacobs, line 21: OTI: I appreciate the small commit size, they don't have to be this small though haha

src/bin/scheduler/setup/directories.rs
SoloJacobs, line 1: I don't know how to comment the commit message, but `be belong` -> `belong` and `these files are anyway cleaned up` -> `these files are cleaned up anyway`.

SoloJacobs: Second point: I'm not sure why it is ok for these permissions to be overwritten. Please explain it in the commit message (also follow-up commit).

SoloJacobs (changes requested):

src/bin/scheduler/setup/directories.rs
jherbel: Good point! I will also check this for our windows permission handling.

src/bin/scheduler/setup/directories.rs
```diff
@@ -427,6 +429,73 @@ fn clean_up_results_directory(
...
+    #[cfg(unix)]
+    fn take_ownership_non_recursive(&self, target: &Utf8Path) -> std::io::Result<()> {
+        std::os::unix::fs::lchown(target, Some(self.user_id), Some(self.group_id))
+    }
+
+    #[cfg(unix)]
+    fn take_ownership_recursive(&self, target: &Utf8Path) -> std::io::Result<()> {
+        self.take_ownership_non_recursive(target)?;
+
+        if target.is_dir() {
```
SoloJacobs, line 528: This poses problems if there are cycles in the directory structure (via symlinks).
```
if target.is_symlink {
  return
  }
```
either before or after `take_ownership_non_recursive`.

SoloJacobs (changes requested):

SoloJacobs:
Nice :+1:

PR #627, opened 2024-11-04
test: run without internet access

SoloJacobs:
The scheduler should run without access to internet (some customer do not want to allow this attack vector).

This is now tested on Linux in a dedicated system test.

CMK-19446

### DIFF
--- a/tests/test_ht_import_scheduler.rs
+++ b/tests/test_ht_import_scheduler.rs
@@ -1,0 +1,205 @@
+#![cfg(unix)]
+pub mod helper;
+use crate::helper::{await_plan_results, directory_entries};
+use anyhow::{bail, Result as AnyhowResult};
+use assert_cmd::cargo::cargo_bin;
+use camino::{Utf8Path, Utf8PathBuf};
+use robotmk::config::{
+    Config, EnvironmentConfig, ExecutionConfig, PlanConfig, PlanMetadata, RCCConfig,
+    RCCEnvironmentConfig, RCCProfileConfig, RetryStrategy, RobotConfig, SequentialPlanGroup,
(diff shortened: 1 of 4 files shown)


tests/test_ht_import_scheduler.rs
```diff
@@ -0,0 +1,208 @@
...
+    n_seconds_run_max: u64,
+) -> AnyhowResult<()> {
+    let config_path = test_dir.join("config.json");
+    write(&config_path, to_string(&config)?)?;
+    let run_flag_path = test_dir.join("run_flag");
+    write(&run_flag_path, "")?;
+
+    let mut robotmk_no_env_cmd = Command::new("sudo");
+    robotmk_no_env_cmd
+        .arg("unshare")
```
jherbel, line 139: Would it make sense to add a small check if internet access as indeed cut, since this is crucial for the test? I was thinking of this:
```
sudo unshare --net -- ping -c 1 github.com && echo "I still have internet access" || [run scheduler]
```

jherbel (changes requested):
Very nice work, only one comment :+1:

tests/test_ht_import_scheduler.rs
SoloJacobs: I have tested just adding the `.arg` as you have suggested here. I think it will just add `&&` as an argument to `ping`, which fails with:
```
ping: invalid option -- '-'
```
(very cryptic).
I think the most straightforward way would be to add a plan, which will fail, without internet. However, as we discussed, this is overkill here.

SoloJacobs: Actually, I think it is ok to just pass a shell binary in the test (inlining the shell script is not as problematic as I thought it would be.) See new change

PR #629, opened 2024-11-04
Change layout of environment building directory

jherbel:


examples/termination/main.rs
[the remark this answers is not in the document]
SoloJacobs: `handling of this runtime directory more similar to the handling of the runtime directory...` reads very weirdly.

src/environment.rs
```diff
@@ -12,7 +12,10 @@ pub enum ResultCode {
 #[derive(Clone, Debug, PartialEq)]
 pub enum Environment {
     System(SystemEnvironment),
-    Rcc(RCCEnvironment),
+    Rcc(
+        // see https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
```
SoloJacobs, line 16: It says always measure the changes this lint suggests ? From what I can tell boxing is a premature optimization here, I would just suppress it.

src/bin/scheduler/setup/general.rs
```diff
@@ -183,6 +187,61 @@ fn setup_rcc_working_directories(
     )
 }
 
+fn setup_environment_building_directories(
```
SoloJacobs, line 190: Why is there so much extra code here? Is this change adding functionality?

examples/termination/main.rs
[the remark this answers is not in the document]
jherbel: Done

src/environment.rs
jherbel: Done

src/bin/scheduler/setup/general.rs
jherbel: Not really, but I found it hard to consolidate this with the already existing functionality. I would suggest that we first make the necessary changes and then check if and how we can reduce code duplication. Ok?

SoloJacobs (approved):
Nice!

src/bin/scheduler/setup/general.rs
SoloJacobs: Yes, that does not bother me. It's not really about the code duplication, but I think we might need better interfaces for adding functionality.

PR #630, opened 2024-11-05
Windows: Do not follow symbolic links when changing the permissions

jherbel:
The only place where we might encounter a symbolic link is the directory where unpack the managed robots to. However, we never change the permissions of the unpacked files. Plus, we certainly don't want to modify the permissions of any files outside our runtime directory.

### DIFF
--- a/src/bin/scheduler/setup/windows_permissions.rs
+++ b/src/bin/scheduler/setup/windows_permissions.rs
@@ -30,7 +30,6 @@ pub fn grant_permissions_to_all_plan_users(
     path: &Utf8Path,
     plans: Vec<Plan>,
     permissions: &str,
-    additional_icacls_args: &[&str],
     description_for_failure_reporting: &str,
 ) -> (Vec<Plan>, Vec<SetupFailure>) {
     let mut surviving_plans = vec![];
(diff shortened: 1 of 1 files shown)


src/bin/scheduler/setup/windows_permissions.rs
```diff
@@ -30,19 +30,21 @@ pub fn grant_permissions_to_all_plan_users(
...
-            let mut icacls_args = vec![path.as_str(), "/grant", &icacls_permission_arg];
-            icacls_args.extend(additional_icacls_args);
-
-            match run_icacls_command(icacls_args).context(format!(
+            match run_icacls_command([
+                path.as_str(),
+                "/grant",
+                &format!("{}:{}", user_session.user_name, permissions),
+                "/T",
+                "/L",
```
SoloJacobs, line 45: Are we sure this works with recursion as expected? Maybe a test would be appropriate here.
```
Performs the operation on a symbolic link instead of its destination.
```

SoloJacobs (approved):
Looks correct to me, but I can't really tell.

jherbel:
Running icacls.exe on a symlink with the /L option activated does not do anything. Based on this experiment, it seems that symlinks don't have their own, separate permissions in Windows. Instead, they always have the same permissions as the target.

Not changing the permissions is unacceptable. Therefore, we run without the /L option and accept that we might modify the permissions of arbitrary files.

PR #634, opened 2024-11-07
Dev/sol/hololib

SoloJacobs:


src/bin/scheduler/setup/rcc.rs
```diff
@@ -319,23 +318,33 @@ fn run_command_spec_once_in_current_session(
     )
 }
 
-fn run_command_spec_per_session(
+fn run_command_spec_per_session<T>(
```
jherbel, line 321: Let's rename the function to
`run_rcc_per_session`
or so

src/bin/scheduler/setup/directories.rs
```diff
@@ -56,18 +56,87 @@ pub fn setup(
     create_dir_all(&global_config.managed_directory)?;
 
     setup_results_directories(global_config, &plans, &ownership_setter)?;
+    #[cfg(windows)]
+    create_dir_all(&global_config.robocorp_home_base)?;
+
     let (surviving_plans, managed_dir_failures) = setup_managed_directories(plans);
+    #[cfg(windows)]
+    let (surviving_plans, robocorp_home_failures) =
+        setup_robocorp_home_directories(global_config, surviving_plans);
```
jherbel, line 65: We should only do this for plans that actually use RCC. I think this should become a step in `setup_rcc_working_directories`. In `setup_rcc_working_directories`, we can do something like
```
let mut rcc_plans_and_rcc_configs = Vec::new();
let mut system_plans = Vec::new();
for plan in plans.into_iter() {
    match plan.environment.clone() {
        Environment::Rcc(rcc_env) => {
            rcc_plans_and_env_build_dirs.push((plan, rcc_env.clone()))
        }
        _ => system_plans.push(plan),
    }
}
```
`rcc_plans_and_rcc_configs` can then be used as input for both `setup_environment_building_directories` and your new function. Ping me in case what I mean is unclear.

src/bin/scheduler/setup/directories.rs
```diff
@@ -56,18 +56,87 @@ pub fn setup(
...
+) -> (Vec<Plan>, Vec<SetupFailure>) {
+    use super::windows_permissions::grant_full_access;
+    use log::info;
+    use robotmk::session::Session;
+
+    let mut surviving_plans = vec![];
+    let mut failures = vec![];
+
+    for (session, plans_in_session) in plans_by_sessions(plans) {
+        let robocorp_home = session.robocorp_home(&global_config.robocorp_home_base);
```
jherbel, line 97: We should take this value from `RCCEnvironment`, not re-compute it. See also `setup_environment_building_directories`.

src/environment.rs
```diff
@@ -94,22 +97,25 @@ impl SystemEnvironment {
 }
 
 impl RCCEnvironment {
-    pub fn bundled_command_spec(binary_path: &Utf8Path) -> CommandSpec {
+    pub fn bundled_command_spec(binary_path: &Utf8Path, robocorp_home: String) -> CommandSpec {
```
jherbel, line 100: Let's make this an instance method instead and use `self.robocorp_home`

src/bin/scheduler/setup/rcc.rs
```diff
@@ -176,8 +165,14 @@ fn enable_long_path_support(
...
     plans: Vec<Plan>,
 ) -> Result<(Vec<Plan>, Vec<SetupFailure>), Cancelled> {
-    let mut command_spec =
-        RCCEnvironment::bundled_command_spec(&global_config.rcc_config.binary_path);
+    use robotmk::session::CurrentSession;
+    let session = Session::Current(CurrentSession {});
+    let mut command_spec = RCCEnvironment::bundled_command_spec(
+        &global_config.rcc_config.binary_path,
+        session
+            .robocorp_home(&global_config.robocorp_home_base)
```
jherbel, line 173: We should use `RCCEnvironment.robocorp_home` here and below. Therefore, we probably need
```
Vec<(Plan, RCCConfig)>
```
as input.

src/session.rs
```diff
@@ -39,6 +39,10 @@ impl Session {
             Self::User(session) => session.id(),
         }
     }
+
+    pub fn robocorp_home(&self, robocorp_home_base: &Utf8Path) -> Utf8PathBuf {
```
jherbel, line 43: Let's remove this, since we hopefully only need to call it one place.

src/bin/scheduler/setup/directories.rs
```diff
@@ -56,18 +56,87 @@ pub fn setup(
     create_dir_all(&global_config.managed_directory)?;
 
     setup_results_directories(global_config, &plans, &ownership_setter)?;
+    #[cfg(windows)]
+    create_dir_all(&global_config.robocorp_home_base)?;
```
jherbel, line 60: We should also take ownership here and below

jherbel (changes requested):

jherbel:
Once we are finished here, we will re-visit our internal datastructures. It seems that we need something like
```
Vec<(Plan, RCCConfig)
```
in multiple places. So it makes sense to at least consider an adjustment here, something `RCCPlan`. We will see.

src/session.rs
SoloJacobs: It's called in three places?

src/environment.rs
SoloJacobs: This is very awkward in those places, which don't have any `RCC` plan (such as turning of telemetry). We would have to make  all fields of `RCCEnvironment` optional. I think some our abstractions do not work.

src/bin/scheduler/setup/directories.rs
SoloJacobs: Hmm, this would require us to call `plans_in_session[0].environment` no? Are we really sure that the plan `RCCEnvironment` is the same environment used to do the initial configuration? I think it is better keep these two constructions independent.

src/bin/scheduler/setup/directories.rs
SoloJacobs: I fully agree that we should not skip the system plans based on what we managed to setup as a robocorp home directory. That was a bug I missed.

However, I think the robocorp home should based on the `Session`, not `RCCEnvironment`. We are not setting up the home for each plan. I am open to an improvement, but this does not seem to be it to me.

src/bin/scheduler/setup/rcc.rs
SoloJacobs: I think this is the third time you mention it in this review. Let's pick one place for the discussion.

jherbel (approved):

jherbel:
As discussed: We will merge this to have the functionality in place. In a follow-up step, we will investigate options to re-work our setup steps. The current data structures are well-suited for environment building and plan scheduling (per-plan operations), but they make for example per-session operations cumbersome.

PR #643, opened 2024-11-19
Migrate RCC setup steps

jherbel:


src/bin/scheduler/setup/rcc.rs
```diff
@@ -122,6 +118,44 @@ impl SetupStep for StepFilePermissions {
     }
 }
 
+struct StepRCCCommand {
+    config: GlobalConfig,
```
SoloJacobs, line 105: Adding `GlobalConfing` as a field is not a good idea. It makes constructing `StepRCCCommand` far more complicated to construct. This is a problem for testing.

src/bin/scheduler/setup/api.rs
```diff
@@ -31,30 +38,33 @@ pub fn skip(plans: Vec<Plan>) -> (Box<dyn SetupStep>, Vec<Plan>) {
...
-                        err.summary,
-                        err.cause,
-                    );
-                    errors.push(SetupFailure {
-                        plan_id: plan.id.clone(),
-                        summary: err.summary.clone(),
-                        details: format!("{:?}", err.cause),
-                    });
+            Err(err) => match err {
+                Terminate::Cancelled => return Err(Cancelled),
```
SoloJacobs, line 50: This feels like it will complicate all the setup steps. Would not be better idea to simply check after each setup step whether we should terminate? A setup step itself still can report whatever it want as an error.

src/bin/scheduler/setup/rcc.rs
```diff
@@ -161,6 +139,74 @@ impl SetupStep for StepRCCCommand {
     }
 }
 
+struct StepDisableSharedHolotree {
+    config: GlobalConfig,
```
SoloJacobs, line 178: As previously mentioned: I don't think adding `GlobalConfig` as a field is a good idea. (But let's discuss this for the other step only).

SoloJacobs (changes requested):

src/bin/scheduler/setup/rcc.rs
jherbel: Done

src/bin/scheduler/setup/api.rs
jherbel: Done

SoloJacobs (approved):
Yay!

PR #645, opened 2024-11-20
Small improvements

jherbel:


src/bin/scheduler/scheduling/scheduler.rs
```diff
@@ -67,7 +69,18 @@ async fn run_sequential_plan_group_scheduler(
             _ = cancellation_token.cancelled() => { return }
         };
         for plan in plans.clone() {
-            let _ = spawn_blocking(move || run_plan(&plan).map_err(log_and_return_error)).await;
+            let plan_id = plan.id.clone();
+            match spawn_blocking(move || run_plan(&plan).map_err(log_and_return_error)).await {
+                Ok(Err(Terminate::Cancelled)) => {
+                    return;
+                }
+                Err(error) => {
```
SoloJacobs, line 77: Does it make sense to add the explicit type in the match expression here? It might future proof us a little against API changes in `spawn_blocking`.

src/bin/scheduler/scheduling/scheduler.rs
```diff
@@ -80,7 +93,13 @@ async fn run_cleanup_job(cancellation_token: CancellationToken, plans: Vec<Plan>
             _ = clock.tick() => { }
             _ = cancellation_token.cancelled() => { return }
         };
-        spawn_blocking(move || cleanup_working_directories(plans.iter()));
```
SoloJacobs, line 83: Wait, why does this not have an await? Did our cleanup job ever work?

src/bin/scheduler/scheduling/scheduler.rs
```diff
@@ -80,7 +93,13 @@ async fn run_cleanup_job(cancellation_token: CancellationToken, plans: Vec<Plan>
...
+        let _ = spawn_blocking(move || cleanup_working_directories(plans.iter()))
+            .await
+            .map_err(|err| {
+                log_and_return_error(
```
SoloJacobs, line 99: Would it not be simpler to log the error without returning it?

src/bin/scheduler/scheduling/scheduler.rs
```diff
@@ -43,7 +45,7 @@ pub async fn run_plans_and_cleanup(global_config: &GlobalConfig, plans: &[Plan])
     info!("Received termination signal while scheduling, waiting for plans to terminate");
     while let Some(outcome) = join_set.join_next().await {
         if let Err(error) = outcome {
-            error!("{error:?}");
+            log_and_return_error(error);
```
SoloJacobs, line 48: What is the difference to just logging the error?

src/bin/scheduler/scheduling/scheduler.rs
jherbel: Yes, they did. You don't have to explicitly await this thread, it runs anyway. Adding the await avoids spawning multiple cleanup threads, but hopefully, this never happened anyway (they shouldn't take 5 minutes).

src/bin/scheduler/scheduling/scheduler.rs
jherbel: In case we decide to log errors differently, this would be taken into account here.

src/bin/scheduler/scheduling/scheduler.rs
jherbel: If we decide to change the logging format in `log_and_return_error`, this would be taken into account here.

src/bin/scheduler/scheduling/scheduler.rs
jherbel: Hmm, how would you do that? Not sure I understand what you mean.

SoloJacobs: Ah, this only works in `python`, I guess? I was thinking along these lines:
```
Err(JoinError(e)) => {
```
But is probably not possible in Rust.

SoloJacobs (approved):

PR #647, opened 2024-11-21
Tighten permissions of ROBOCORP_HOME base (Windows only)

jherbel:


tests/test_scheduler.rs
```diff
@@ -660,3 +666,35 @@ fn assert_sequentiality(
...
+        [
+            "current_user",
+            #[cfg(windows)]
+            &format!("user_{headed_user_name}"),
+        ]
+    );
+    #[cfg(windows)]
+    {
+        let permissions_robocorp_home_base = get_permissions(robocorp_home_base).await?;
+        assert_eq!(
```
SoloJacobs, line 686: It  does not really feel this is testing something?

SoloJacobs (approved):
Approve with nit

tests/test_scheduler.rs
jherbel: It seems quite tricky to test localization-independent that only the administrator group has access. Any suggestions?

SoloJacobs: From a quick look, it appears `/findsid` and `/q` are relevant parameters to `icacls`. Otherwise, we could consider switching to the `win32` API or removing the check altogether.

jherbel: /findsid is what I needed. /q does not seem to have any effect on my VM, see also the comment in the code

SoloJacobs (approved):

PR #653, opened 2024-11-25
Offer direct download links for RCC in release page

jherbel:
CMK-20341

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,7 +33,14 @@ jobs:
           name: rmk_linux64
           path: artifact/rmk_linux64/
 
-      - run: zip -r executables.zip artifact
+      - run: zip -r all_executables.zip artifact
+
+      - name: "Prepare RCC binaries for release"
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -54,7 +61,13 @@ jobs:
       - uses: ncipollo/release-action@v1
         with:
           allowUpdates: false
-          artifacts: "executables.zip,assets/robotmk_core-1.0.0.mkp"
+          artifacts: >
+            executables.zip,
```
SoloJacobs, line 65: The name `executables.zip` seems inconsistent now? Should we not change this?

SoloJacobs (changes requested):

.github/workflows/release.yaml
jherbel: What would you change it to? "all_executables" or so? It still contains the same files as before.

jherbel: Done

SoloJacobs (approved):

PR #654, opened 2024-11-25
Implement option to configure secret environment variables for RF executions

jherbel:


src/command_spec.rs
```diff
@@ -15,7 +15,7 @@ impl Display for CommandSpec {
         let env_str = self
             .envs
             .iter()
-            .map(|(k, _)| format!("{k}=***"))
```
SoloJacobs, line 18: This should probably be reviewed by the security team instead?

src/command_spec.rs
```diff
@@ -65,8 +65,11 @@ impl CommandSpec {
         self
     }
 
-    pub fn add_env(&mut self, key: String, value: String) -> &mut Self {
-        self.envs.push((key, value));
+    pub fn add_env<T>(&mut self, key: T, value: T) -> &mut Self
```
SoloJacobs, line 75: FYI: This was deliberate by me. Functions, which offer passing by reference, which immediately clone the value, are considered non-idiomatic by some. I guess we need to agree on a consistent style at some point.

src/config.rs
```diff
@@ -78,6 +78,7 @@ pub struct RobotConfig {
     pub variable_files: Vec<Utf8PathBuf>,
     pub argument_files: Vec<Utf8PathBuf>,
     pub exit_on_failure: bool,
+    pub secret_environment_variables: Vec<RobotFrameworkSecretEnvVar>,
```
SoloJacobs, line 81: How can these be secret, if they are written to the config? Wouldn't it be more sensible to add them to `robot.yaml`? What exactly does secret mean here?

SoloJacobs (changes requested):

src/command_spec.rs
jherbel: I don't have a definitive opinion on this. I would say it is a case-by-case decision. In this case, I oriented myself towards the standard library: https://doc.rust-lang.org/nightly/std/process/struct.Command.html#method.env

src/config.rs
jherbel: Secret means that they are obfuscated in the Checkmk UI. But they are written to the config in plain text. Any agent plugin that needs a secret does the same. I am not aware of any better solution at the moment. Users can of course hard-code them in their test definitions if they want to.

src/command_spec.rs
jherbel: Why? Generally, it is OK to log environment variables. The ones which specifically should not be logged should go to the `secret_env_vars` field. But we certainly want to log for example `ROBOCORP_HOME` or `RCC_REMOTE`.

src/config.rs
SoloJacobs: Ack

SoloJacobs (commented):
I still think, I can't review this change.

src/command_spec.rs
jherbel: done

SoloJacobs (approved):

PR #655, opened 2024-11-26
Reset RCC binary and profile permissions during setup

jherbel:
Since explicitly grant plan users access to these files, we should also reset the access to cover the case where a plan user is removed or changes.

### DIFF
--- a/src/bin/scheduler/setup/steps/rcc.rs
+++ b/src/bin/scheduler/setup/steps/rcc.rs
@@ -6,7 +6,7 @@ use super::{
 use crate::internal_config::{GlobalConfig, Plan};
 use crate::logging::log_and_return_error;
 #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


SoloJacobs (commented):
I don't see how a user having read/execute permissions for RCC is an issue? Any user with an internet connection could execute such a file, no? Or am I missing something Windows specific?

jherbel:
It is not really an issue, but since we explicitly grant these permissions, we should also remove them I would say.

SoloJacobs:
Hmm, but it does not seem correct to me to skip the rcc plans in case the fails, since this step does not seem necessary. Having unecessary steps is a little odd.

jherbel:
We decided to approach this differently: Instead of granting access to specific users, we will grant access to the Users group. This way, we won't have to deal with resetting permissions.

PR #675, opened 2025-02-07
Reduce number of dynamically linked dependencies for Windows binaries

jherbel:


jherbel:
The customer responded, compiling Robotmk this way fixes the issue.

rust-toolchain.toml
```diff
@@ -0,0 +1,11 @@
...
+# actions-rust-lang/setup-rust-toolchain when setting up the CI environment.
+
+[toolchain]
+channel = "1.83"
+
+[toolchain.linux]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[toolchain.windows]
+targets = ["x86_64-pc-windows-gnu"]
```
SoloJacobs: Should this not be x86_64-pc-windows-msvc?

.github/workflows/robotmk_build.yaml
```diff
@@ -10,8 +10,6 @@ jobs:
       - uses: actions/checkout@v4
```
SoloJacobs: You wrote "warnins" in the commit message.

jherbel: Done

rust-toolchain.toml
jherbel: Not in this change, this happens in the subsequent commit

SoloJacobs (approved):

PR #678, opened 2025-02-19
Reduce number of dynamically linked dependencies for Windows binaries (3.0)

jherbel:


SoloJacobs:
I think it would nice to mention pros and cons for the musl target. But OTI.

SoloJacobs (approved):

PR #679, opened 2025-02-19
Remove targets from `rust-toolchain.toml` and re-add to Rust setup step

jherbel:
The documentation of `actions-rust-lang/setup-rust-toolchain` says: "First, all items specified in the toolchain file are installed. Afterward, the components and target specified via inputs are installed in addition to the items from the toolchain file."

However, this does not seem to be the case. The targets specified in `rust-toolchain.toml` are apparently ignored. This makes not difference in the 2.0 branch, since we use the default Linux/Windows targets. But in 3.0 and main, we want to use `x86_64-unknown-linux-musl`, which is not the default Linux target. Specifying this target in `rust-toolchain.toml` does not make `actions-rust-lang/setup-rust-toolchain` install it, which is how this problem became apparent.

Even though we are currently using the default Linux/Windows targets, we still specify them in the Rust setup step for the sake of explicitness.

### DIFF
--- a/.github/workflows/robotmk_build.yaml
+++ b/.github/workflows/robotmk_build.yaml
@@ -11,6 +11,7 @@ jobs:
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1.5.0
         with:
+          target: x86_64-pc-windows-msvc
           # By default, setup-rust-toolchain sets "-D warnings". As a side effect, the settings in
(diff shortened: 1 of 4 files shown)


SoloJacobs (approved):

SoloJacobs:
The documentation is misspelling `Afterwards` haha.

PR #697, opened 2025-05-12
Include micromamba binaries in releases

jherbel:


SoloJacobs (dismissed):

jherbel:
Turned out that micromamba is quite particular regarding the filename of its own executable

PR #708, opened 2025-05-26
Add dedicated test for packed Conda environments

jherbel:
This test runs the scheduler without internet connection to ensure that packed environments can be used on offline test nodes.

CMK-23556

### DIFF
--- a/tests/minimal_suite/conda.yaml
+++ b/tests/minimal_suite/conda.yaml
@@ -2,7 +2,8 @@ channels:
   - conda-forge
 
 dependencies:
-  - python=3.11.6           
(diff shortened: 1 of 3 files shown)


tests/minimal_suite/conda.yaml
```diff
@@ -2,7 +2,8 @@ channels:
   - conda-forge
 
 dependencies:
-  - python=3.11.6           
-  - pip=23.3.1 
+  - python=3.11.6
+  - pip=23.3.1
+  - conda-pack=0.8.1
```
SoloJacobs, line 7: Why is this necessary?

PR #710, opened 2025-05-28
Do not remove all plan working directories at startup

jherbel:
Instead, only remove the working directories of plans that are no longer configured. The working directories of configured plans are cleaned up regularly and shouldn't be removed at startup.

SUP-23960


SoloJacobs (changes requested):
This definitely needs a test to document why have this, and which files are "user-facing".

SoloJacobs (approved):

PR #790, opened 2026-04-13
Update dependencies in RF suite used in tests

jherbel:


jherbel:
Backported from main, skipping review

